PR #23, opened 2023-04-01
Gatsby build testing

pmelab:
## Description of changes

* Run a Gatsby build during the integration test phase.
* Switch to hoisted node linking.

## Motivation and context

* Make sure on every change that the Gatsby build still works.
* This surfaced and issue with Gatsby partial hydration and transitive dependencies in pnpm. The latter contain `+` characters in their real system paths, which are replace with " " (spaces) in webpack. That in turn causes the bundler to fail. Switching to hoisted node linking resolved the problem, but compromises on some packaging guarantees that pnpm provides.

## How has this been tested?

- [x] Manually
- [ ] Unit tests
- [x] Integration tests


Leksat:
Looks like everything in here is already outdated 😅 4 days is too much for this project 😂

PR #26, opened 2023-04-04
ci: switch back to standard node linker

pmelab:
## How has this been tested?

- [ ] Manually
- [ ] Unit tests
- [ ] Integration tests


pmelab:
Does not work for now, since gatsby won't work with the pnpm linker and `deploy` does not work with the node linker. We will go for `turbo prune` instead.

PR #33, opened 2023-04-11
GraphQL schema tests

Leksat:
## How has this been tested?

- [ ] Manually
- [ ] Unit tests
- [ ] Integration tests


Leksat:
Using Playwright is an overhead  here. But it was so tempting to use it because everything was already set up.

~Unfortunately~ Luckily Playwright adds platform name to the snapshots, so we can't really use it.

PR #47, opened 2023-04-18
GraphQL snapshot tests

Leksat:
## How has this been tested?

- [ ] Manually
- [ ] Unit tests
- [x] Integration tests

### DIFF
--- a/tests/schema/specs/content.spec.ts
+++ b/tests/schema/specs/content.spec.ts
@@ -1,0 +1,168 @@
+import gql from 'noop-tag';
+import { expect, test } from 'vitest';
+
+import { fetch } from './lib';
+
+test('Page minimal', async () => {
+  const result = await fetch(gql`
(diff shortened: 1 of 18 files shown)


tests/schema/specs/content.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+    ",
+            },
+          ],
+        },
+      },
+    }
+  `);
+});
+
+// TODO: Implement after images are integrated.
```
Leksat, line 107: @pmelab FYI: I was not able to add default content for `BlockImage` because `gatsby build` fails in this case. It tries to json-parse an object in https://github.com/AmazeeLabs/silverback-mono/blob/162f2ce48966f695b14e135a92478df1d4104782/packages/npm/%40amazeelabs/scalars/src/image.tsx#L34

PR #66, opened 2023-05-10
fix: add check for pageContext before calling locale

dan2k3k4:


dan2k3k4:
This might not be needed, as I had removed `SilverbackPageContext` from the private frontend UI assessment repo which perhaps requires `pageContext.locale` to exist.

Leksat:
#thanksdan

PR #68, opened 2023-05-10
fix: remove line-clamp as now part of default for tailwind

dan2k3k4:


dan2k3k4:
As per warning, removed from config:

```shell
@custom/ui:prep:styles: warn - As of Tailwind CSS v3.3, the `@tailwindcss/line-clamp` plugin is now included by default.
@custom/ui:prep:styles: warn - Remove it from the `plugins` array in your configuration to eliminate this warning.
```

pmelab (changes requested):
I suppose we can also remove the `@tailwindcss/line-clamp` package then?

PR #69, opened 2023-05-11
feat: strangler edge function

pmelab:
## Description of changes

Add an edge function that allows to selectively pass requests to other systems. This can be used to proxy specific pages to legacy systems, proxy specific requests directly to Drupal (Webforms?) or let Drupal handle unknown redirects.

## How has this been tested?

- [x] Manually
- [ ] Unit tests
- [x] Integration tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,23 +129,8 @@ The template includes a Netlify Edge Function
 unknown requests selectively to other systems. This can be used to replace only
 specific pages of a legacy system or incorporate existing business logic.
 
-At the top of `strangler.ts`, you will find an array definition of `LegacyHost`
-objects. These have the following properties:
-
-- **`url`**: The base url of the legacy system.
-- **`urlFilter`** (optional): A function that inspects a given `URL` object and
-  determines if it should be processed with this edge function or not.
-- **`responseFilter`** (optional): A function that inspects a given response and
-  decides if it should be ahnded back to the client.
-
-The "strangler" will first check if a given path is already present in the
-Netlify build. If thats the case, the Netlify site takes precedence and no
-proxying happens. Otherwise, it will iterate through the defined legacy systems,
-and the response of the first system where `urlFilter` and `responseFilter` both
-are undefined or return `true` will be returned to the client. If no legacy
-system was able to handle the request, the standard `404` page will be returned.
-
-By default, the application includes one legacy system, which is Drupal. All
-unknown requests are proxied to Drupal, but only `301` and `302` responses are
-taken into account. That way, redirects can be handled by Drupal, but no other
-routes are exposed via Netlify.
+Refer to the
+[Strangler Pattern](https://www.martinfowler.com/bliki/StranglerFigApplication.html)
+blog post if you wonder where the name comes from, to
+[Edge functions documentation](https://docs.netlify.com/edge-functions/overview/)
+for technical details and to `strangler.ts` for how to add new legacy systems.
--- a/README.md
+++ b/README.md
@@ -121,3 +121,31 @@ In a standard project we use three fixed Lagoon environments: `dev`, `stage` and
 Lagoon should also be configured to create automatic environments for feature
 branches that are prefixed with `lagoon/`. Those will be filled with test
 content.
+
+## "Strangling" legacy systems
+
+The template includes a Netlify Edge Function
+(`apps/website/netlify/edge-functions/strangler.ts`) that allows to proxy
+unknown requests selectively to other systems. This can be used to replace only
+specific pages of a legacy system or incorporate existing business logic.
+
+At the top of `strangler.ts`, you will find an array definition of `LegacyHost`
+objects. These have the following properties:
+
+- **`url`**: The base url of the legacy system.
+- **`urlFilter`** (optional): A function that inspects a given `URL` object and
+  determines if it should be processed with this edge function or not.
+- **`responseFilter`** (optional): A function that inspects a given response and
+  decides if it should be ahnded back to the client.
+
+The "strangler" will first check if a given path is already present in the
+Netlify build. If thats the case, the Netlify site takes precedence and no
+proxying happens. Otherwise, it will iterate through the defined legacy systems,
+and the response of the first system where `urlFilter` and `responseFilter` both
+are undefined or return `true` will be returned to the client. If no legacy
+system was able to handle the request, the standard `404` page will be returned.
+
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -121,3 +121,31 @@ In a standard project we use three fixed Lagoon environments: `dev`, `stage` and
...
+specific pages of a legacy system or incorporate existing business logic.
+
+At the top of `strangler.ts`, you will find an array definition of `LegacyHost`
+objects. These have the following properties:
+
+- **`url`**: The base url of the legacy system.
+- **`urlFilter`** (optional): A function that inspects a given `URL` object and
+  determines if it should be processed with this edge function or not.
+- **`responseFilter`** (optional): A function that inspects a given response and
+  decides if it should be ahnded back to the client.
```
Leksat: ahnded -> handed ?

apps/website/netlify/edge-functions/homepage-redirect.ts
```diff
@@ -1,13 +1,12 @@
-// @ts-ignore
-import Negotiator from 'https://deno.land/x/negotiator/mod.ts';
+import Negotiator from 'https://deno.land/x/negotiator@1.0.1/mod.ts';
+import { Config } from 'https://edge.netlify.com';
```
Leksat, line 2: What are the best practices for Deno imports? Should we always pin a certain version?

Or `https://edge.netlify.com` is a special case and it should always be the latest?

README.md
```diff
@@ -121,3 +121,31 @@ In a standard project we use three fixed Lagoon environments: `dev`, `stage` and
 Lagoon should also be configured to create automatic environments for feature
 branches that are prefixed with `lagoon/`. Those will be filled with test
 content.
+
+## "Strangling" legacy systems
```
Leksat, line 125: I would give a very very brief explanation here and move the rest of the docs to `strangler.ts` directly. Reasons:
- There the docs will really help with development
- Less chances that docs will be outdated

apps/website/netlify/edge-functions/strangler.ts
```diff
@@ -0,0 +1,85 @@
+import { Config, Context } from 'https://edge.netlify.com';
+
+type LegacyHost = {
+  url: string;
+  urlFilter?: (url: URL) => boolean;
```
Leksat: How about making it more abvious?
```ts
applies: (url: URL) => boolean;
```
And then we set it to `applies: () => true` in the drupal host?

apps/website/netlify/edge-functions/strangler.ts
```diff
@@ -0,0 +1,85 @@
...
+  responseFilter?: (response: Response) => boolean;
```
Leksat: How do you think, is there a chance that we will want to modify the `response`? If yess, maybe it should return `Response | null`?

apps/website/netlify/edge-functions/strangler.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  // Otherwise, pass the request to the legacy applications.
+  for (const legacyUrl of legacySystems) {
+    const request = originalRequest.clone();
+    const targetUrl = new URL(legacyUrl.url);
+    // Check if we even want to proxy this request.
+    // Skip if the urlFilter exists and returns false.
+    if (legacyUrl.urlFilter && !legacyUrl.urlFilter(targetUrl)) {
+      console.log('Skipping', legacyUrl.url);
```
Leksat: Will we see this in the Publisher logs? If so, maybe let's prefix it with `[strangler] `? (Same for other `console.log`s)

apps/website/netlify/edge-functions/strangler.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export const config = {
+  path: '/*',
+  excludedPath: [
+    '/__preview/*',
+    '/_gatsby/*',
+    '/page-data/*',
+    '/static/*',
+    '/*.js',
+    // TODO: add more paths that we know are served statically for sure or are not legacy.
```
Leksat, line 97: The `/sites/default/files/*` redirect defined in `gatsby-node.ts` - does it take precedence?

tests/e2e/specs/redirects.spec.ts
```diff
@@ -0,0 +1,13 @@
...
+
+import { websiteUrl } from '../helpers/url';
+
+test.describe('drupal redirects', () => {
+  test('are detected', async ({ page }) => {
+    await page.goto(websiteUrl('/node/1'));
+    const content = await page.getByRole('main');
+    await expect(
+      content.getByText('The requested page is currently in the bathroom.'),
+    ).not.toBeVisible();
```
Leksat: Maybe test the final URL? E.g. starts with `/en/` and does not contain `/node/1`

Leksat (approved):
Question and nitpicks. Great job in general!

apps/website/netlify/edge-functions/strangler.ts
pmelab: Thats actually a leftover 😬

apps/website/netlify/edge-functions/strangler.ts
pmelab: We can add `/sites/default/files` here, since we know its handled by a rewrite 🖊️

tests/e2e/specs/redirects.spec.ts
pmelab: 👍

apps/website/netlify/edge-functions/strangler.ts
Leksat: But if it's visible in the log, that can be very good help for debugging

apps/website/netlify/edge-functions/strangler.ts
Leksat: +1

apps/website/netlify/edge-functions/homepage-redirect.ts
pmelab: I think in that case it should be latest, because in the netlify runtime we also have no control over the version.

apps/website/netlify/edge-functions/strangler.ts
pmelab: But its also a lot of logs.

Leksat: True, let's remove them

PR #74, opened 2023-05-25
Content-hub

pmelab:
## Description of changes

Example implementation of a content hub.

## Motivation and context

To cover technological requirements for dynamic listings and forms.

## How has this been tested?

- [x] Manually
- [x] Unit tests
- [x] Integration tests


packages/drupal/custom/src/Plugin/GraphQL/Directive/ContentHub.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+class ContentHub extends PluginBase implements DirectiveInterface {
+
+  public function buildResolver(ResolverBuilder $builder, array $arguments): ResolverInterface {
+    return $builder->callback(function ($parent, $args) {
+      // TODO: Switch this to views.
+      $offset = $args['pagination']['offset'];
+      $limit = $args['pagination']['limit'];
+
+      $countQuery = \Drupal::entityQuery('node');
```
Leksat, line 26: Maybe like this?
```
$query = ...
$count = (clone $query)->count()->execute();
$pageIds = $query->...
```
To not duplicate the query setup.

packages/schema/src/schema.graphqls
```diff
@@ -43,10 +57,20 @@ type FooterNavigation implements Navigation @menu(menu_id: "footer") {
...
+  @type(id: "page") {
   locale: Locale! @resolveEntityLanguage
-  path: String! @resolveEntityPath @isPath
+  path: Url! @resolveEntityPath @isPath
   title: String! @resolveProperty(path: "title.value")
+  teaserImage: MediaImage
+    @resolveEditorBlocks(
+      path: "body.value"
+      aggregated: ["core/paragraph", "core/list"]
+      ignored: ["core/group"]
```
Leksat, line 70: If we need just the hero image, do we still need to aggregate and ignore anything?

packages/schema/src/schema.graphqls
```diff
@@ -90,7 +114,14 @@ type BlockImage @type(id: "drupalmedia/drupal-media-entity") {
...
 
+input PaginationInput {
+  limit: Int!
+  offset: Int!
+}
+
 type Query {
   previewPage(id: ID!, rid: ID, locale: String!): Page
     @fetchEntity(type: "node", id: "$id", rid: "$rid", language: "$locale")
+  contentHub(query: String, pagination: PaginationInput!): ContentHubResult!
```
Leksat, line 125: I recently use a single `input` parameter
```gql
type Query {
  contentHub(input: ContentHubInput!): ContentHubResult!
}

input ContentHubInput {
  query: String
  pagination: PaginationInput!
}
```
I just like to have a separate type for it. Even if I don't actually use it :)

tests/schema/specs/content-hub.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+      {
+        contentHub(pagination: { limit: 2, offset: 0 }) {
+          items {
+            title
+          }
+          total
+        }
+      }
+    `);
+    expect(result.data.contentHub.total).toBe(12);
```
Leksat, line 18:
```
.toBeGreaterThan(2)
```
To edit this test less in the future.

tests/schema/specs/content-hub.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+      {
+        contentHub(query: "Arch", pagination: { limit: 3, offset: 0 }) {
+          total
+          items {
+            path
+          }
+        }
+      }
+    `);
+    expect(result.data.contentHub.items[0].path).toBe('/en/architecture');
```
Leksat, line 64: How do we search in German? 🤔 

Also, `/content-hub` page does not have a language prefix at the moment. And `/en/content-hub` is 404 😬

packages/ui/src/components/Molecules/Pagination.tsx
```diff
@@ -0,0 +1,81 @@
...
+  ArrowLongRightIcon,
+} from '@heroicons/react/20/solid';
+import clsx from 'clsx';
+import React from 'react';
+import { z } from 'zod';
+
+import { useIntl } from '../../utils/intl';
+
+export const paginationParamsSchema = z.object({
+  page: z.coerce.number().default(1),
```
Leksat, line 13: I recently also add `catch()` to each param with a default value. Because `zod` is more about validation than parsing. Currently `/content-hub?page=foo` leads to WSOD.

packages/ui/src/components/Molecules/Pagination.tsx
```diff
@@ -0,0 +1,81 @@
...
+});
+
+export function useCurrentPage(): number {
+  const [location] = useLocation();
+  return paginationParamsSchema.parse(
+    Object.fromEntries(new URLSearchParams(location.search).entries()),
```
Leksat, line 19: I used `query-string` instead of `URLSearchParams`. Because it supports arrays by default.

Also, here is what I came up with for arrays in URL params:
```ts
const arrayOfStrings = z
  .array(z.string())
  .optional()
  .catch((ctx) => (typeof ctx.input === 'string' ? [ctx.input] : undefined));
```

tests/e2e/specs/content-hub.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    await expect(content.getByText('Technologies')).toBeVisible();
+  });
+
+  test('allows to search for items', async ({ page }) => {
+    await page.goto(websiteUrl('/content-hub'));
+    const content = await page.getByRole('main');
+    content.getByPlaceholder('Keyword').fill('technologies');
+    content.getByRole('button', { name: 'Search' }).click();
+    await expect(content.getByText('Architecture')).not.toBeVisible();
+    await expect(content.getByText('Technologies')).toBeVisible();
```
Leksat, line 29: In a real project I would create a dedicated content with `ContentHubTest: ...` titles, and then always search with `ContentHubTest` keyword. So that new default content don't break tests.

apps/website/src/pages/content-hub.tsx
```diff
@@ -0,0 +1,16 @@
+import { ContentHub } from '@custom/ui/routes/ContentHub';
+import React from 'react';
+
+import { Wrapper } from '../utils/wrapper';
+
+export function Head() {
+  return <meta title="Page not found" />;
```
Leksat, line 7: What is it for?

Leksat (commented):
Looks great!

Thanks for the separate commits 💚
(I also try to do that, but mine are muuuuch bigger 😅)

Issue #52, opened 2023-04-20
Interface translations


pmelab:
- [ ] import translations into drupal
- [ ] source them via gatsby

PR #76, opened 2023-05-27
Interface translations

chindris:
## Description of changes

Feeds the react intl provider with translations from the CMS

## Related issues
https://github.com/AmazeeLabs/silverback-template/issues/52

## How has this been tested?

- [ ] Manually
- [ ] Unit tests
- [x] Integration tests (there's still a todo left)


chindris:
@pmelab There are a couple of integration tests for this, but we actually need to have a test database with translations added. Previously we used to have a install-cache zip which we could update and it was used for such cases. Not sure what is the workflow now, and how we'd handle this case. We could of course have some kind of an import script, like we have for default content, is that what we should do?

pmelab:
this is solved slightly different in #139

PR #79, opened 2023-06-03
Strangler on Lagoon

pmelab:
## Description of changes

Refactor the strangler mechanism to a serverless function that runs after all other Netlify path resolutions (statically built files and rewrites/redirects).

## Motivation and context

This approach solves several problems:

* It will never run on resources in the static site, since they have a higher precedence now.
* ... therefore it does not need to know if the request is part of the static site.
* ... therefore no complicated exclusions necessary.
* It has a lower precedence than any redirects, which allows more control over what handles which requests.
* It does not need to pass through requests to Netlify which was error prone.
* Function code can be shared as a package.

## How has this been tested?

- [x] Manually
- [ ] Unit tests
- [x] Integration tests


apps/website/netlify/functions/strangler.ts
```diff
@@ -0,0 +1,18 @@
+import { createStrangler } from '@amazeelabs/strangler-netlify';
+import fs from 'fs';
+
+export const handler = createStrangler(
+  [
+    {
+      // For Standard drupal redirects, we are interested in any
+      // url (therefore no urlFilter) and only in redirects (301, 302).
+      url:
+        process.env.DRUPAL_INTERNAL_URL ||
```
Leksat, line 10: Where are the serverless functions executed? On Netlify? If yes, then `DRUPAL_INTERNAL_URL` should be unreachable... Or do I miss something?

Leksat (commented):
Looks great!

PR #83, opened 2023-06-16
Fixes and tests for Gutenberg blocks

Leksat:
## Description of changes

Content editing:
- `custom/quote` block is replaced with `core/quote`
- Video cannot be added to `custom/hero` anymore

Development:
- New `turbo:dev:light` script for `website` app. It does the same thing as `turbo:dev`, but does not rebuild `cms`, it just starts it. I could not come up with a better script name 😅 
- `save: () => <InnerBlocks.Content />` was added to all blocks as a [default](https://github.com/AmazeeLabs/idmc/pull/38#pullrequestreview-1466664225)
- Added `skipTypename: true` to the codegen config
  - Previous behavior: Always add `__typename` as optional to all types
  - Now: Add `__typename` as a required param to types only if it was explicitly set in the query (operations)

Frontend:
- `h2` is changed to `h1` in `custom/hero`

## Motivation and context

It was found that some blocks just do not work. So I wrote tests and added fixes.

## How has this been tested?

- [ ] Manually
- [ ] Unit tests
- [x] Integration tests


pmelab (approved):
😮 Thats a lot. Thanks!

PR #86, opened 2023-06-28
feat(publisher): authenticate with drupal

colorfield:
## Description of changes

Keep only project specific configuration to integrate https://github.com/AmazeeLabs/silverback-mono/pull/1339

SLB-176

## Motivation and context

Better DX

## How has this been tested?

- [x] Manually
- [ ] Unit tests
- [ ] Integration tests


apps/cms/config/sync/user.role.administrator.yml
```diff
@@ -42,6 +44,7 @@ permissions:
   - 'dropzone upload files'
   - 'edit config_pages entity'
   - 'translate editable entities'
+  - 'trigger a gatsby build'
```
colorfield: Added `trigger a gatsby build` and `access publisher` permissions to the administrator role as these ones are going in pairs (if we access publisher, we can trigger a build). But we need to check if it's fine for this role.

apps/cms/scaffold/settings.php.append.txt
```diff
@@ -1,5 +1,5 @@
 // Basic settings
-$settings['hash_salt'] = getenv('DRUPAL_HASH_SALT') ?: 'banana123';
+$settings['hash_salt'] = getenv('DRUPAL_HASH_SALT') ?: 'time-flies-like-an-arrow-fruit-flies-like-a-banana';
```
Leksat, line 2: We need to either revert this or adjust it in the init script https://github.com/AmazeeLabs/silverback-template/blob/9084adacbe1c031461f22696e951b5e3c05aa0d3/INIT.md?plain=1#L89-L93

apps/website/publisher.config.ts
```diff
@@ -36,4 +36,15 @@ export default defineConfig({
   },
   databaseUrl: 'persisted-store/publisher.sqlite',
   publisherPort: isLagoon ? 3000 : 8000,
+  oAuth2: {
```
Leksat: Do I get it right that oauth will be enabled by default on all new projects? And that `Publisher authentication with Drupal` section is a must-do for every new project? (Or every environment?)

If so, I'd make it as short as possible, and automate it as much as possible. For example, it says to adjust `/admin/config/people/simple_oauth`, but as far as I can see, the correct values are already set in the Drupal config. (Am I wrong?)

The general idea is:
- Someone creates a new project following the [instructions](https://github.com/AmazeeLabs/silverback-template/blob/dev/README.md#create-a-new-project-from-this-template)
- Most of things are [automated](https://github.com/AmazeeLabs/silverback-template/blob/dev/INIT.md)
- There are minimum required steps to do
- There are no unclear steps to do

(The `Environment overrides` section is on my list to make it smaller/shorter. But I haven't had time for this yet.)

Leksat (commented):
I updated all the things (https://github.com/AmazeeLabs/silverback-template/pull/89) so there are conflicts now. Sorry 😅

apps/cms/scaffold/settings.php.append.txt
colorfield: We need at least 32 chars otherwise the OAuth library throws. So the `randomString(32)` is good / I will replace the init, thanks for spotting!

apps/website/publisher.config.ts
colorfield: Yes, I was also thinking that it could be more simple for project specific things.

- For this `/admin/config/people/simple_oauth` most of the config is done, we just need to create the certificates per project
- For `/admin/config/services/consumer` that's content entities / has to be done per environment
- Environment variables could also be automated

I will check what could already be done with the init script.

Leksat: Oh, btw, the code blocks in the init script are marked as `ts`, but in fact it's JavaScript. TypeScript is not yet supported by `@amazeelabs/mzx` 🙈

apps/cms/scaffold/settings.php.append.txt
colorfield: fixed in https://github.com/AmazeeLabs/silverback-template/pull/86/commits/c71a69e6d4f815aaad5c78369870ed07e2e25192

apps/website/publisher.config.ts
colorfield: @Leksat i finally disabled it by default, everywhere. This might be better DX and so we can enable it only when requested + it won't complain when creating a new environment if something is not set.

As discussed: locally we are using `isLagoon` and skip in any case. I think there is no use case for that for now, and if we want to debug it locally it should be a task to be done in the `silverback-mono` Publisher package.

For Lagoon environments: we set `PUBLISHER_SKIP_AUTHENTICATION=true` by default and documentation was reviewed to make the setup clearer.

apps/cms/config/sync/user.role.publisher.yml
```diff
@@ -0,0 +1,9 @@
+uuid: a6375354-2d5f-4632-b53e-619040dec631
+langcode: en
+status: true
+dependencies: {  }
+id: publisher
+label: Publisher
+weight: -5
+is_admin: null
+permissions: {  }
```
Leksat: What is this role for? It has no permissions, so looks suspicious.

README.md
```diff
@@ -165,6 +165,72 @@ lagoon runtime configuration.
...
+  via Drush: cd in the cms directory then
+  
+  ```bash
+  drush simple-oauth:generate-keys ./keys
+  ```
+  
+  or via the UI
+  
+  - Go to `/admin/config/people/simple_oauth`
+  - Click on "Generate keys", the directory should be set to `../keys`
```
Leksat: I think we changed this to private files?

Leksat (commented):
Two questions + `hook_file_download` and we are good to go 🚀

apps/cms/config/sync/user.role.publisher.yml
colorfield: It is only used for the Consumer (content entity) scope.
As we are just granting access to Publisher currently, it has no permissions.

It could be used as a placeholder role, for what the Publisher UI could do with Drupal without being current user specific. I cannot see a use case for now so we might just remove it.

colorfield: And also adjust the Consumer creation in the deploy hook.

README.md
colorfield: Thanks, fixed https://github.com/AmazeeLabs/silverback-template/pull/86/commits/b47b2c45ae2782d108e0b57266bb5fa29c4be314

apps/cms/config/sync/user.role.publisher.yml
Leksat: The role can confuse someone... like me 😅 

As an admin, I'd like to know what is this role for and when to assign it to users. Could you answer this?

colorfield: Sorry for the late reply.
The original reasoning behind is similar to the `Gatsby Build` role with the `Gatsby` user. It's a way to limit how Publisher can interact with Drupal with this Consumer.

But as 
- we don't have any `Publisher` user currently
- we have another authentication system for Gatsby to source from Drupal
- there is nothing specific that `Publisher` can do as we are using the current user permissions check - specifically the `access publisher` permission, e.g. for the `administrator` role here https://github.com/AmazeeLabs/silverback-template/pull/86/files/f3fecfcbe3eb2e600b744bb4785377c0209b3842#diff-dda0826d23d77d3c618b19281ec0a74bb7d14e8f13e6e85d021bb0f77c1b1a2eR32

let's just remove it as it can also be confusing / we might be tempted to assign this role to users instead of the permission.

colorfield: Removed here
- https://github.com/AmazeeLabs/silverback-template/pull/86/commits/96fa474761657c49cb408c4888e6dc4ac1208541
- and not using it as part of the Consumer setup https://github.com/AmazeeLabs/silverback-template/pull/86/commits/6e1fc325392c86cf78886a8e5ef6903f4269e6f5#diff-36927e5378571a60a719b3904a4d26d00bcc8137c4d512d2cd4e9f01dcfbc686R47-R53 - as opposed to https://github.com/AmazeeLabs/silverback-mono/blob/development/packages/composer/amazeelabs/silverback_gatsby/modules/silverback_gatsby_oauth/silverback_gatsby_oauth.deploy.php#L20-L23 and https://github.com/AmazeeLabs/silverback-mono/blob/development/packages/composer/amazeelabs/silverback_gatsby/modules/silverback_gatsby_oauth/silverback_gatsby_oauth.deploy.php#L59-L61

Also moving Consumers setup from `silverback-mono` https://github.com/AmazeeLabs/silverback-mono/pull/1448 to the `custom` module as it's template specific.

packages/drupal/custom/custom.module
```diff
@@ -180,3 +181,14 @@ function _custom_key_auth_form_access(UserInterface $user): AccessResult {
...
   return $access;
 }
+
+/**
+ * Implements hook_file_download().
+ *
+ * Prevent any access to keys.
+ */
+function custom_file_download($uri) {
+  if (str_starts_with($uri, 'private://') && str_ends_with($uri, '.key')) {
```
Leksat, line 191: I think `private://.key?foo=bar` can trick this check.

Leksat (approved):
Please merge after improving `custom_file_download` 🙏 🎉

packages/drupal/custom/custom.module
colorfield: In `hook_file_download($uri)`, `$uri` is not sensitive to query string so should be fine.

PR #115, opened 2023-11-20
Swappable operation implementations

pmelab:
## Description of changes

- Introduce a new system for swapping execution of GraphQL operations. The new shared package `@amazeelabs/executors` allows to register execution functions for specific GraphQL operations and parameter combinations.
- Unify all data-access to use `useOperation` and this new system under the hood.

> Note: The package is currently part of the PR, but will be moved to `silverback-mono` before merging.

## Motivation and context

This approach solves two problems:

1. Take out the guesswork "where" data comes from in UI components. Right now, it depends on the implementation. Gatsby data comes from props, dynamic queries from `useOperation`. This binds the UI to the implementation and makes it harder to use components in other environments (e.g. previews). Now every data operation has a GraphQL operation in the schema package and the UI component always uses `useOperation` and the implementation (Gatsby, Decap, Next ...) controls how the query is resolved.
2. Make it easier to swap out data sources for different purposes, even based on parameter matching (e.g. pre-render page 1 of content hub, resolve others with http-request to Drupal).

Use cases:

- preview handling (drupal, decap ...)
- simplify data-mocking in storybook
- pre-rendering of listings
- injection of global data (menus, breadcrumbs) without [prop-drilling](https://kentcdodds.com/blog/prop-drilling)

## How has this been tested?

- [x] Manually
- [x] Unit tests
- [x] Integration tests


packages/executors/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "prep": "tsc",
+    "build": "tsc",
+    "test:unit": "vitest run",
+    "test:static": "tsc --noEmit && eslint \"**/*.{ts,tsx,js,jsx}\" --ignore-path=\"./.gitignore\" --fix"
+  },
+  "keywords": [],
+  "author": "Amazee Labs <development@amazeelabs.com>",
+  "license": "ISC",
+  "dependencies": {
+    "@types/lodash-es": "^4.17.12",
```
Leksat: `devDependencies`?

packages/executors/src/lib.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+export function registerExecutor(executor: Executor): void;
+
+export function registerExecutor(id: string, executor: Executor): void;
+
+export function registerExecutor(
+  id: string,
+  variables: VariablesMatcher,
+  executor: Executor,
```
Leksat: TS function overloading does not always play well. For example, guess the inferred types:
```ts
type FirstParam = Parameters<typeof registerExecutor>[0];
type LastParam = Parameters<typeof registerExecutor>[2];
```
[Answer](https://www.typescriptlang.org/play?ssl=22&ssc=57&pln=21&pc=1#code/C4TwDgpgBAogHhAxgV2AewE5QLwFgBQUUAPlAIYB2IBRpAFHQJYAmAXFAM7AaMUDmAGigA3MjzIAjADYQO7AEpJMzADxce-IZRAA+AJQ4d5KiSgAFDGgC2jDhBXb9AbgIFQkKADUxjSTI4AsmTAiAAWEFh4hKaKiMpq3LyCxro0pgyiGHJQsfHqSVpU+oZQEmhoMpR6LvgEEHBgmMBQAGbIFIjAjGgUUBgQfLbAEfBIqJh09WPoGOyjKDN67MJoLDV1DU2t7Z3dvf2DXCMICxMs7PmaUFOns7An4xhLIqvM6-j1jRjNbR1dPX0BkNjtMJmlzpxEpo0plfNJZOxvOJ4YFgmEIgI0jdHnMHjNMfhnis1q4Pptvts-ntAYdhhh5o86AA6FliPjZdoAawoaAA7hQANoAXQMAG8AL6k9zQABijCywDMYjIVhw5mVVggdI4KmlaBaNOB9LxmB0AoADEKatKoAAZMhcJUYFVqp0qrURHV6g0HI0MmZmgBMVqAA)

I would suggest using a single object parameter:
```ts
export function registerExecutor(args: RegistryEntry) {
  registry.push(args);
}
```

IMHO it would make make the function calls easier to read:
```ts
registerExecutor({ executor: myFunc, id: 'foo-bar' });
```

---

Or am I missing something?

packages/executors/src/lib.ts
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Create an executor from the registry.
+ *
+ * @param id
+ *   The operation ID, from the schema package.
+ * @param variables
+ *   A dictionary of variables to be passed to the operation.
+ */
+export function createExecutor(id: string, variables?: Record<string, any>) {
```
Leksat: What's the return type?

Leksat: BTW, in my pet projects I use `'@typescript-eslint/explicit-function-return-type': 'error'` eslint rule. But not sure if it will play well at work 🤔

Leksat (approved):
Looks great in general!

packages/executors/src/lib.ts
pmelab: I thought it would produce a more natural API:
```javascript
// register an executor that matches everything
registerExecutor(iDoEverythingOverHttp);
// register one for a specific operation;
registerExecutor(MyOperationId, myOperationImplementation);
// for a specific set of input:
registerExecutor(ContentHubQuery, {page: 1}, staticFirstPage);
```
Compared to:
```javascript
// register an executor that matches everything
registerExecutor({executor: iDoEverythingOverHttp});
// register one for a specific operation;
registerExecutor({id: MyOperationId, executor: myOperationImplementation});
// for a specific set of input:
registerExecutor({id:ContentHubQuery, args: {page: 1}, executor: staticFirstPage});
```

I'm not sure if it has to be possible to infer types here, since the`Operation*` type helpers are available to infer return types and variables:
```typescript
type first = AnyOperationId;
type second = Partial<OperationVariables<first>>;
type third = () => Promise<OperationResult<first>>;
```

But I also don't have a very strong opinion on this. How strong are yours? Should we pull in more voices?

packages/executors/src/lib.ts
pmelab: Thats a great discussion point. Explicit means a type error is found earlier, not potentially through three levels of generic functions that infer something.

PR #116, opened 2023-12-11
SLB-207: Fix Gatsby builds on lagoon

pmelab:
## Description of changes

Move composer install up into the `builder` layer of the Dockerfile, so composer dependencies are available durin `prep` and the full schema definition with all directive definitions can be assembled.
## Motivation and context

Gatsby builds on lagoon where broken due to missing directive definitions.
## How has this been tested?

- [x] Manually
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/.lagoon/Dockerfile
+++ b/.lagoon/Dockerfile
@@ -1,7 +1,12 @@
 # ====================================================================================================
 # BUILDER IMAGE
 # ====================================================================================================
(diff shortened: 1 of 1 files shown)


Leksat (changes requested):
Mua-ha-ha-ha-ha!

Leksat (approved):
Ah-ah-ah-ah-aum!

PR #126, opened 2024-01-11
SLB-202: De-duplicate Gatsby queries

pmelab:
## Description of changes

* Introduce a new package `@amazeelabs/gatsby-plugin-operations` that modifies Gatsby in way
that allows us to use the same Operation-ID's we have everywhere else. (to be moved to `silverback-mono` after review)
* Unify handling of Decap- and Drupal pages (does not use `@isPath` any more, since its Drupal-only)
* Refactor all Gatsby queries to use documents from `@custom/schema`
* Remove the `gatsby-fragments.js` artefact, since its not required any more.

## Motivation and context

* De-duplicate query operations (SLB-202)
* Full type safety in Gatsby
* Simpler Gatsby templates and static queries
* Full control over Gatsby page creation process
  * Allows custom queries (e.g. additional context data)
  * Less magic

## How has this been tested?

- [x] Manually
- [x] Unit tests
- [x] Integration tests


packages/gatsby-plugin-operations/src/graphql.ts
```diff
@@ -0,0 +1,30 @@
...
+  _graphql = graphql;
+  _operations = JSON.parse(readFileSync(operations).toString());
+}
+
+/**
+ * Execute a graphql query against gatsby.
+ */
+export function graphqlQuery(id: string, vars?: any): any {
+  if (!_graphql) {
+    throw '_graphql has not been initialized';
```
Leksat: Here and following: `throw 'string'` can lead to weird behavior of debug tools. I'd recommend `throw new Error('string')`.

Or was there a specific reason to throw a string?

packages/gatsby-plugin-operations/src/plugin.ts
```diff
@@ -0,0 +1,64 @@
...
+                ),
+              ),
+            );
+            path.skip();
+            return;
+          }
+        }
+      },
+    },
+  }) as PluginObj<
```
Leksat: `satisfies`?

packages/gatsby-plugin-operations/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
```
Leksat: Where did you get all of these settings? Just curious.

My latest strategy for `tsconfig.json`: start with empty, only add settings which are really required.

packages/schema/src/schema.graphql
```diff
@@ -67,8 +65,9 @@ type FooterNavigation implements Navigation @menu(menu_id: "footer") {
   items: [NavigationItem]! @lang @resolveMenuItems
 }
 
-interface Page implements ContentHubResultItem {
+interface Page implements ContentHubResultItem @value(json: "\"page\"") {
```
Leksat: What's the reason for `@value` here?

apps/website/src/pages/404.tsx
```diff
@@ -1,36 +1,25 @@
...
-  }
-`;
+export const query = graphql(NotFoundPageQuery);
 
 function isTruthy<T>(value: T | undefined | null): value is T {
   return Boolean(value);
 }
 
-export default function Index({ data }: PageProps<NotFoundPageQuery>) {
+export default function Index({ data }: PageProps<typeof query>) {
```
Leksat, line 22: Niiiice 😁

apps/website/gatsby-node.mjs
```diff
@@ -17,11 +17,20 @@ export const onCreateWebpackConfig = ({ actions }) => {
   });
 };
 
+/**
+ * @template T extends any
+ * @param {T | undefined | null} val
+ * @returns {val is T}
```
Leksat, line 23: Wow! Did not know it's possible :mindblown:

Leksat (approved):
Awesome work 💚

apps/website/gatsby-node.mjs
pmelab: Yeah, its awesome! Perfect compromise between no type checking and the catastrophy of typescript in gatsby-node.

apps/website/src/pages/404.tsx
pmelab: 💪

packages/schema/src/schema.graphql
pmelab: Drupal needs a value to resolve the `Page` interface. Even though there is only one subtype within Drupal. I changed it to use the entity bundle. Thats more clear and practical (I hope).

packages/gatsby-plugin-operations/tsconfig.json
pmelab: Copy & Paste. I reviewed them a little.

packages/gatsby-plugin-operations/src/graphql.ts
pmelab: I took the opportunity and improved the error message.

PR #128, opened 2024-01-15
Webform styling

Leksat:
## Motivation and context

https://github.com/AmazeeLabs/silverback-mono/issues/1347

## How has this been tested?

- [x] Manually
- [ ] Unit tests
- [ ] Integration tests

There are integration tests in https://github.com/AmazeeLabs/silverback-mono/pull/1484

Also: Chromatic 🎉


tests/e2e/webform-snapshots/webforms.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  await page.getByLabel('Email with a confirmation').fill('asd@asd');
+  await page.getByRole('button', { name: 'Submit' }).click();
+  await expect(
+    page.locator('[data-drupal-messages] [role="alert"]'),
+  ).toBeVisible();
+  await savePage(page, 'error');
+});
+
+async function savePage(page: Page, name: string) {
```
pmelab, line 25: Should we move `savePage` and the `ScraperPlugin` to a shared package? Maybe in a later step.

pmelab (approved):
Awesome work!

One thing that came into my mind: This also allows us to storybook-ify pure Drupal projects! 🤯

tests/e2e/webform-snapshots/webforms.spec.ts
Leksat: It uses `Page` type which comes from `@playwright/test`. If we move it to a package, we can get issues with Playwright versions. So I would not.

Leksat: Or we change it to
```ts
async function savePage(url: string, content: string, name: string) {
```
and call it as
```ts
await savePage(page.url(), await page.content(), 'name');
```
but then it's not nice anymore 😅

pmelab: TBH, when thinking about testability and reusability, that makes a lot of sense 🤔 
We could use that scraper for other tasks, outside of playwright as well.

Leksat: https://www.npmjs.com/package/@amazeelabs/save-webpage

PR #129, opened 2024-01-22
SLB-201: remove copy operations and streamline artefacts

pmelab:
## Description of changes

* extract static file handling in Gatsby into `@amazeelabs/gatsby-plugin-static-dirs` (to be moved to `silverback-mono` after review)
* rename Drupal stylesheets to make them easier to understand
* remove alls cases where `prep` steps generate files _outside_ of the current page, and make the consuming packages load them via `node_modules` instead.

## Motivation and context

* Reduce development hiccups when forgetting to re-run prep in another package to copy the right files.
* Reduce problems with Turborepo caches (🤞)
* Easier understanding where some files come from.

## How has this been tested?

- [x] Manually
- [ ] Unit tests
- [x] Integration tests


packages/gatsby-plugin-static-dirs/gatsby-node.js
```diff
@@ -0,0 +1,35 @@
+import { cpSync, rmSync } from 'fs';
```
Leksat: `rmSync` isn't used.

Yes, I am a linter 😅

apps/website/gatsby-ssr.ts
```diff
@@ -27,4 +27,11 @@ export const onRenderBody: GatsbySSR['onRenderBody'] = ({
       // We don't know the language.
     }
   }
+  setHeadComponents([
+    React.createElement('link', {
+      key: 'styles.css',
+      rel: 'stylesheet',
+      href: '/styles.css',
```
Leksat: I assume this will introduce an issue with browser cache.

Would it be possible to change it to something like `/styles.css?hash={hash of styles.css contents}`?

Leksat (changes requested):
I'd do something with the css browser cache issue.

The rest looks awesome!

Also, could you assign it to me for a manual testing once it is deployed?

apps/website/gatsby-ssr.ts
Leksat: Hm. Actually this is probably a great improvement to the Netlify deploy time. Now Gatsby won't need to change all html files on any css change. Right?

packages/gatsby-plugin-static-dirs/gatsby-node.js
pmelab: I didn't add eslint to the package! 😆

apps/website/gatsby-ssr.ts
pmelab: > I assume this will introduce an issue with browser cache.

Damn, thats really good catch.

> Hm. Actually this is probably a great improvement to the Netlify deploy time. Now Gatsby won't need to change all html files on any css change. Right?

In theory, yes. But css changes based on classes used in components, and if components change, it re-renders. But maybe we have less to upload.

pmelab: We can't do the hash, because the file is not yet there at that point. But i added a timestamp-cachebuster.

pmelab: Just realized that this means a new stylesheet bust on every content change. Will think more about it 😬

Leksat: Ah, wait, if we add content hash to the css url, there is no difference at all 😅

Leksat: > Just realized that this means a new stylesheet bust on every content change

A change to css url would re-upload all files to Netlify 🙈

Leksat: > We can't do the hash, because the file is not yet there at that point.

It's in the node_modules, no?

Leksat: Or! Could it be that I am completely wrong and https://www.npmjs.com/package/serve-static solves all the browser cache issues with magical headers?

pmelab: `serve-static` is only there during `gatsby dev`. 
Just tried to create the hash, and Gatsby says "unsafe usage of `readFileSync`, i'm going to rebuild EVERYTHING EVERY TIME".
Also we have a lot of other assets (html files), that update correctly. I think netlify takes care of that.

Leksat: Let's try without hash 🤞 I will test in all major browsers.

pmelab: The hash is back baby!
Using Gatsby standard behavior now. See commit messages.

pmelab: If we have component- and therefore stylesheet-change, we have to re-render and re-upload everything anyway. But having non-inline stylesheets makes more sense with tailwind.

apps/website/gatsby-browser.ts
```diff
@@ -1,3 +1,5 @@
+import './node_modules/@custom/ui/build/styles.css';
```
Leksat: I saw that previously we imported it from both gatsby-browser and gatsby-ssr. Now it's gatsby-browser only. Not a mistake?

apps/website/gatsby-config.mjs
```diff
@@ -38,6 +38,7 @@ export default {
     siteUrl: process.env.NETLIFY_URL,
   },
   plugins: [
+    'gatsby-plugin-uninline-styles',
```
Leksat, line 41: This plugin does the same thing you did in the beginning? 🤔

Leksat: Ah, the filename already contains the hash. I see.

Leksat (approved):
Looks good! Let's test on Prod 😁

apps/website/gatsby-browser.ts
pmelab: It was a mistake before.

PR #135, opened 2024-01-30
SLB-217: simplify data access

pmelab:
## Description of changes

Change the way pages are loaded to only use the path and `useOperation`. No data is passed through props and no id's are used.

Handling of 404 and Home pages as well as localization moved into the UI package and is framework-agnostic now as well. Everything is based on standard graphql queries.

`@amazeelabs/react-intl-rsc` has been replaced with `react-intl`, which works with server components by now. We don't need the wrapper any more.

## Motivation and context

Truly decouple the UI package. All data is fetched using `useOperation` now, and no framework specific arguments are required. The same component would seamlessly work in a different React environment, and it should be easier to understand whats happening.

The UI component now retrieves information from `useLocation` and `useOperation` without making any assumptions on where the data comes from. Gatsby transparently provides implementations for these.

Example:
The `ViewPage` operation ...
https://github.com/AmazeeLabs/silverback-template/blob/11c4b5058aa9777c694b8578b5a2bde5861d0972/packages/schema/src/operations/ViewPage.gql#L1-L5

... is executed with the current path as argument in the route of the UI package ...

https://github.com/AmazeeLabs/silverback-template/blob/11c4b5058aa9777c694b8578b5a2bde5861d0972/packages/ui/src/components/Routes/Page.tsx#L10-L11

... and pre-filled by Gatsby in the respective template:

https://github.com/AmazeeLabs/silverback-template/blob/11c4b5058aa9777c694b8578b5a2bde5861d0972/apps/website/src/templates/page.tsx#L42-L44

## How has this been tested?

- [x] Manually
- [x] Unit tests
- [x] Integration tests


packages/schema/src/schema.graphql
```diff
@@ -10,6 +10,11 @@ implementation(gatsby): ./image.js#imageProps
 """
 directive @imageProps repeatable on FIELD_DEFINITION | SCALAR | UNION | ENUM | INTERFACE | OBJECT
 
+"""
+Retrieve a page from the Gatsby datastore by its path. foo
```
Leksat: `foo` - did you mean `Woo hoo`? 😛

packages/schema/src/schema.graphql
```diff
@@ -173,9 +178,10 @@ type Query {
 
   allPages: [Page] @gatsbyNodes(type: "Page")
   websiteSettings: WebsiteSettings
-  viewPage(id: ID!, rid: ID, locale: String!): Page
-    @fetchEntity(type: "node", id: "$id", rid: "$rid", language: "$locale")
-    @gatsbyNode(type: "Page", id: "$id")
+  viewPage(path: String!): Page
+    @route(path: "$path")
+    @fetchEntity
+    @gatsbyPageByPath(path: "$path")
```
Leksat: Do I understand it right?

In Drupal context it will use `@route` and `@fetchEntity`. And `@gatsbyPageByPath` will be ignored.

In Gatsby it will be the opposite - only `@gatsbyPageByPath` will be used.

apps/website/src/templates/page.tsx
```diff
@@ -39,19 +34,12 @@ export function Head({ data }: HeadProps<typeof query>) {
...
-  );
-  return (
-    <Page id={pageContext.id} locale={(pageContext.locale || 'en') as Locale} />
-  );
+export default function PageTemplate({ data }: PageProps<typeof query>) {
+  // Retrieve the current location and prefill the
+  // "ViewPageQuery" with these arguments.
+  // That makes shure the `useOperation(ViewPageQuery, ...)` with this
+  // path immediately returns this data.
+  const [loc] = useLocation();
```
Leksat: Why `loc` and not `location`?

I can read `loc` as location, local, localhost, locked and even Loki 😄

Leksat (approved):
OMG, It's better and better with every new PR 😄

packages/schema/src/schema.graphql
pmelab: Exactly. Every system uses the directives it knows. This might be confusing at some point, but i have no better idea (yet).

packages/schema/src/schema.graphql
pmelab: 😆 I have no clue.

pmelab: Sorry, cloo-hoo.

packages/schema/src/schema.graphql
pmelab: Technically one directive can be implemented in multiple systems 🤔  I have to think about that.

pmelab: @Leksat: I pushed another experimental commit. Instead of re-using the `@route` and `@fetchEntity` directive, i moved that code into a custom `@pageByPath` directive, that works both in Drupal and Gatsby.

I'm not sure if thats preferable or not:

**Pro:**
* its easy to understand where things are implemented
* its easy to modify the implementation

**Con:**
* it duplicates code from the two directives

pmelab: **Semi-pro:** its a nice example of a directive implemented in multiple systems

packages/drupal/custom/src/Page.php
Leksat, line 1: I see some advanced logic which includes redirects, access checks, cache tags... It lives in the custom module and has no test coverage. This feels dangerous.

Is it possible to make the service reuse `@route` and `@fetchEntity` directives internally?

packages/schema/src/page.ts
```diff
@@ -1,13 +1,44 @@
...
   undefined,
   { nodeModel: any },
   { path: string }
-> = async (_, { path }, { nodeModel }) => {
-  return await nodeModel.findOne({
-    type: 'Page',
-    query: { filter: { path: { eq: path } } },
-  });
+> = (_, { path }) => {
+  return new URL(path, process.env.NETLIFY_URL || 'https://localhost:8000');
```
Leksat: Let's wrap with a try-catch?

Leksat (commented):
Almost there 😄

packages/schema/src/page.ts
pmelab: Done!

PR #137, opened 2024-02-06
SLB-260: split prep tasks

pmelab:
## Description of changes

Remove all occurences of `&&` in `prep` steps and replace them with Turborepo chains.

## Motivation and context

Consistent approaches for multistep builds and small performance gains due to more granular caching.

## How has this been tested?

- [ ] Manually
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/packages/schema/turbo.json
+++ b/packages/schema/turbo.json
@@ -1,19 +1,52 @@
 {
   "$schema": "https://turborepo.org/schema.json",
-  "extends": ["//"],
+  "extends": [
+    "//"
+  ],
   "pipeline": {
-    "prep": {
-      "dependsOn": ["@custom/cms#prep:composer"],
+    "prep:codegen": {
+      "dependsOn": [
+        "@custom/cms#prep:composer"
+      ],
       "inputs": [
-        "../../apps/website/node_modules/@amazeelabs/*/directives.graphql",
-        "../../apps/cms/web/modules/contrib/*/directives.graphql",
-        "../../apps/cms/web/modules/custom/*/directives.graphql",
-        "src/**",
         "codegen.ts",
         ".graphqlrc.json",
-        "tsconfig.json"
+        "src/**/*.gql",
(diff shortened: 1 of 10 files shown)


packages/schema/turbo.json
```diff
@@ -1,19 +1,52 @@
...
-      "dependsOn": ["@custom/cms#prep:composer"],
+    "prep:codegen": {
+      "dependsOn": [
+        "@custom/cms#prep:composer"
+      ],
       "inputs": [
-        "../../apps/website/node_modules/@amazeelabs/*/directives.graphql",
-        "../../apps/cms/web/modules/contrib/*/directives.graphql",
-        "../../apps/cms/web/modules/custom/*/directives.graphql",
-        "src/**",
```
Leksat, line 11: Aren't we still depending on these external schema bits?

Leksat (commented):
We still have a pstorm watcher for `schema` package:
https://github.com/AmazeeLabs/silverback-template/blob/409dbca2e7bbdb2df13da67ee2c450e285f3f896/.idea/watcherTasks.xml#L5

packages/schema/turbo.json
pmelab: Good question. Yes, we do. But they are outputs of `@custom/cms#prep:composer`, which we depend on here. So that should be fine, no?

pmelab: I think this is a leftover from times where composer was not covered by turbo.

Leksat: We'll see if it works :)

pmelab: Actually not true! 😄 
For composer stuff, maybe, but directives in custom modules still need to be depended on.

pmelab: Which should be in the central `schema.graphql` anyway ...

Leksat (approved):

PR #138, opened 2024-02-07
feat(SLB-194): quick start gutenberg block generation script

nbaosullivan:
## Description of changes

Add a script and package.json command to easily generated Gutenberg block files based off a schema type. Currently has basica support but really does speed up my day-to-day workflow.

## Motivation and context

Speed up development time by automating repetitive tasks. Developers manually add to index.ts in gutenberg_blocks but this could also be automated.

## How has this been tested?

Locally. Example output can be seen in demo-block.tsx

### DIFF
--- a/packages/drupal/gutenberg_blocks/package.json
+++ b/packages/drupal/gutenberg_blocks/package.json
@@ -6,7 +6,7 @@
     "dev": "vite build --watch",
     "prep": "vite build",
     "test:static": "tsc --noEmit && eslint \"**/*.{ts,tsx,js,jsx}\" --ignore-path=\"./.eslintignore\"",
-    "generate-block": "node ./scripts/generate-gutenberg-block.js DemoBlock"
+    "gutenberg:generate": "node ./scripts/generate-gutenberg-block.js"
   },
   "dependencies": {
     "@types/node": "^18",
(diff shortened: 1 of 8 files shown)


pmelab (changes requested):
Thanks! 🥳  This script almost feels big enough to move it to a shared cli tool. But maybe we do that at a later stage when it has been used more.

But we definitely need a documentation block in the README.md!

packages/drupal/gutenberg_blocks/package.json
```diff
@@ -5,10 +5,14 @@
   "scripts": {
     "dev": "vite build --watch",
     "prep": "vite build && cp node_modules/@custom/ui/build/gutenberg.css css/ui.css",
-    "test:static": "tsc --noEmit && eslint \"**/*.{ts,tsx,js,jsx}\" --ignore-path=\"./.eslintignore\""
+    "test:static": "tsc --noEmit && eslint \"**/*.{ts,tsx,js,jsx}\" --ignore-path=\"./.eslintignore\"",
+    "generate-block": "node ./scripts/generate-gutenberg-block.js DemoBlock"
```
pmelab: Scripts can take arguments. That way it could actually be used.

nbaosullivan: Thanks, updated this now.

pmelab (approved):

PR #139, opened 2024-02-13
SLB-269: finish multilingual features

pmelab:
## Description of changes

* Make content hub react to the current language
* Add a simple language switcher
* Add interface translations to decap
* Add interface translations to drupal
* Integration tests for all of these

**Note:** This contains copies for files from https://github.com/AmazeeLabs/silverback-mono/pull/1494. As soon as it is merged, they will be removed from here.

## Motivation and context

Make the starter kit fully multilingual. With Drupal as well as Decap

## How has this been tested?

- [x] Manually
- [x] Unit tests
- [x] Integration tests


apps/cms/prep-database.sh
```diff
@@ -1,13 +1,13 @@
...
-then
+if ! test -f web/sites/default/files/.sqlite; then
   pnpm drupal-install
   pnpm export-webforms
-else
-  pnpm drush cr
 fi
+
+# In any case, re-import translation string.
+pnpm import-translations
```
Leksat, line 12: Should we import translations during the Lagoon deployments?

Leksat (commented):
I think the UI Test failure is a real one :)

apps/cms/prep-database.sh
pmelab: I think so? Unless i'm misunderstanding something 😬 
When frontend code changes, there might be new translations and we have to import them.

Leksat: I mean, this file isn't executed in Lagoon context. And I haven't seen `import-translations` anywhere else in the codebase. So perhaps we need to add something to Dockerfile?

pmelab: Ah, no. This has already been there before. Its in a post-deploy step: https://github.com/AmazeeLabs/silverback-template/blob/b5c153a384218b611d112b3eef69d93d893b8afb/.lagoon.yml#L22

Leksat: Ah!

pmelab:
> I think the UI Test failure is a real one :)

It is not 😛

Leksat (approved):

PR #140, opened 2024-02-14
SLB-273 Check code formatting in CI

Leksat:
## Motivation and context

Developers using auto-formatting-on-save produce unrelated changes.

## How has this been tested?

It was tested on a private project. 8 month, 4 developers. All happy.


package.json
```diff
@@ -10,6 +10,9 @@
   "scripts": {
     "postinstall": "husky install || true",
     "commit": "git-cz",
+    "test:format": "pnpm test:format:root && pnpm test:format:workspaces",
+    "test:format:root": "pnpm prettier --check '**/*.{js,cjs,mjs,ts,jsx,tsx,gql,graphql,graphqls,md,mdx,json}' --ignore-path='./.prettierignore'",
+    "test:format:workspaces": "pnpm --workspace-concurrency=1 -r exec prettier --check '**/*.{js,cjs,mjs,ts,jsx,tsx,gql,graphql,graphqls,md,mdx,json}' --ignore-path='./.gitignore'",
```
pmelab, line 15: Not that it really matters, but why restrict concurrency?

pmelab (approved):

package.json
Leksat: Prettier output isn't recognized by pnpm 🤷 If running in parallel, the output is a mess. You have a list of file paths relative to the workspace roots in random order and there is no chance to guess which file belongs where - the workspaces paths are printed at the bottom.

PR #147, opened 2024-03-05
Mutation example

chindris:


apps/website/src/templates/contact.tsx
```diff
@@ -0,0 +1,6 @@
+import { Contact } from '@custom/ui/routes/Contact';
+import React from 'react';
+
+export default function ContentHubPage() {
```
pmelab: Copy & Paste! 😈

packages/drupal/custom/src/Contact.php
```diff
@@ -0,0 +1,118 @@
...
+    $this->webformService = $webformService;
+  }
+
+  /**
+   * Creates a contact webform entry.
+   */
+  public function create(DirectiveArguments $args): array {
+    try {
+      $submissionData = [
+        'name' => $args->args['contact']['name'],
```
pmelab: Can we generalize this service to take in arbitrary data? Storing input in webform submissions is a very common case, and then we would not need a separate service and directive for each and could move that into the shared Drupal module!

packages/schema/src/schema.graphql
```diff
@@ -16,6 +16,11 @@ implementation(drupal): custom.webform::url
 """
 directive @webformIdToUrl(id: String!) on FIELD_DEFINITION
 
+"""
+implementation(drupal): custom.contact::create
+"""
+directive @createContact(contact: ContactInput) on FIELD_DEFINITION
```
pmelab: To generalize this, we could just remove the argument, and make the directive simply grab the first field argument.

packages/ui/src/components/Molecules/ContactForm.stories.ts
```diff
@@ -0,0 +1,9 @@
+import { Meta, StoryObj } from '@storybook/react';
+
+import { ContactForm as Component } from './ContactForm';
+
+export default {
+  component: Component,
+} satisfies Meta<typeof Component>;
+
+export const Empty = {} satisfies StoryObj<typeof Component>;
```
pmelab: This would be a good opportunity to add [play functions](https://storybook.js.org/docs/writing-stories/play-function) to:

* Test for visual regressions in the filled form
* Test for visual regressions in error display
* Test field validation in general.

Also it is possible to mock executors of the mutation to have stories that show success and error messages. The [Content Hub stories](https://github.com/AmazeeLabs/silverback-template/blob/release/packages/ui/src/components/Organisms/ContentHub.stories.tsx) are a good example for this.

apps/website/src/utils/drupal-executor.ts
```diff
@@ -9,23 +9,52 @@ export function drupalExecutor(endpoint: string, forward: boolean = true) {
...
-              'SLB-Forwarded-Proto': window.location.protocol.slice(0, -1),
-              'SLB-Forwarded-Host': window.location.hostname,
-              'SLB-Forwarded-Port': window.location.port,
-            }
-          : {},
-      })
-    ).json();
-    if (errors) {
-      throw errors;
+    if (variables && variables.graphqlOperationType === 'mutation') {
```
pmelab: This feels like "additional knowledge" to me, that all other executors then would have to implement as well. I think we should derive from "OperationId". I think for now it would be good enough to search for `Mutation:` in `id`. And there is [SLB-259](https://amazeelabs.atlassian.net/browse/SLB-259) which could make this more solid in a later step.

pmelab (changes requested):

pmelab:
One general thing i forgot during the review: We should also add an E2E test to the whole process.

apps/website/src/templates/contact.tsx
chindris: should be fixed now.

packages/drupal/custom/src/Contact.php
chindris: this should be fixed now. The Contact class does not exist anymore and the Webform class is able to create submissions to any webform. The data to be submitted is json encoded.

packages/schema/src/schema.graphql
chindris: this should be fixed. There is a directive now called createSubmission() which can create submissions to any webforms.

packages/ui/src/components/Molecules/ContactForm.stories.ts
chindris: the tests are there, the storybook play functions are working well, except the case when we need the mocked executors (the part that you want to refactor in SLB).

apps/website/src/utils/drupal-executor.ts
chindris: this should be fixed, the check uses the operation id now.

chindris:
> One general thing i forgot during the review: We should also add an E2E test to the whole process.

e2e tests for creating a submission, as well as for validating a wrong e-mail are implemented now.

pmelab (approved):

PR #151, opened 2024-03-13
SLB-204: token-based decap authentication

pmelab:
## Description of changes

* Introduce a new `@amazeelabs/token-auth-middleware` package for authenticating requests.
* Introduce a new `@amazeelabs/decap-cms-backend-token-auth` backend for Decap CMS.
* Swap out `git-gateway` with these two packages.

## Motivation and context

Remove the dependency on Netlify's expensive and inflexible "Identity" feature.

## How has this been tested?

- [x] Manually
- [x] Unit tests
- [x] Integration tests

## Steps to be done after review

- [x] Move `@amazeelabs/token-auth-middleware` to `amazeelabs/silverback-mono`
- [x] Move `@amazeelabs/decap-cms-backend-token-auth` to `amazeelabs/silverback-mono`
- [x] Set Decap's target branch in `main.tsx` back to `prod`

In the packages to move:
- [x] Remove turbo files
- [x] Adjust scripts (`prep` to `build`)


packages/token-auth-middleware/README.md
```diff
@@ -0,0 +1,110 @@
...
+};
+```
+
+By default, the middleware does not need any deeper integration. It works purely
+on HTTP. When accessing an restricted resource, the middleware will redirect to
+a generic login form, which will prompt the user to authenticate. Then
+middleware will then send a login link to the user, which will grant access to
+the resource.
+
+There is a very simple Javascript library that allows to implement this process
```
Leksat: Which library?

packages/token-auth-middleware/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+Backends are responsible for identifying users and sending login links. The
+packages provides a `PostmarkEmailBackend` which can use the postmark messaging
+service to send login links. It inherits from the abstract `EmailBackend` class,
+which can be used to implement email-based authentication with other services.
+
+Email backends allow to restrict access to specific addresses and domains.
+
+```typescript
+const backend = new TestEmailBackend({
```
Leksat: Maybe `PostmarkEmailBackend`? It is mentioned in the paragraph above.

packages/token-auth-middleware/src/lib/handler.ts
```diff
@@ -0,0 +1,282 @@
...
+
+export type TokenAuthHandlerOptions = {
+  /**
+   * Custom login page to redirect to.
+   */
+  customLogin?: string;
+
+  /**
+   * Lifetime of the login token in seconds.
+   * Defaults to one minute.
```
Leksat: Belinda said it took 90 seconds for her to get the email. One minute is too little I guess. Let's make it 5?

packages/token-auth-middleware/src/lib/handler.ts
```diff
@@ -0,0 +1,282 @@
...
+  /**
+   * @throws TokenExpiredError
+   * @throws TokenInvalidError
+   * @param token
+   */
+  validate(token: string): Promise<string>;
+}
+
+export interface AuthenticationBackendInterface<TInfo extends Info> {
+  getInfo(id: string): Promise<TInfo | undefined>;
```
Leksat, line 57: I was wondering how the returned info can be used and found that the `___status` endpoint returns it to the user. Should we document this somehow? That the info should only include the public information.

packages/token-auth-middleware/src/lib/handler.ts
```diff
@@ -0,0 +1,282 @@
...
+  deliver(id: string, link: string): Promise<void>;
+}
+
+export function metaRedirect(message: string, destination: string) {
+  return `<html lang="en"><head><title>${message}</title><meta http-equiv="refresh" content="0;url=${destination}" /></head></html>`;
```
Leksat: I can't come up with an exploit, but still I think we should escape both params.

packages/token-auth-middleware/src/lib/handler.ts
```diff
@@ -0,0 +1,282 @@
...
+}
+
+export function cookieHeader(token: string, path: string) {
+  return `session=${token}; HttpOnly; Secure; SameSite=Strict; Path=${path}/`;
```
Leksat: Why the trailing slash?

packages/decap-cms-backend-token-auth/src/backend.ts
```diff
@@ -0,0 +1,197 @@
...
+   */
+  authComponent() {
+    return AuthComponent;
+  }
+
+  /**
+   * Destroy the current user session.
+   */
+  async logout() {
+    await this.client.logout();
```
Leksat, line 80: Should we call `logout` or something on `this.delegate`?

packages/token-auth-middleware/src/lib/handler.ts
pmelab: What do you mean by "public"? Status only returns values to the session owner.

Leksat: True, not public. Accessible to the user.

packages/token-auth-middleware/src/lib/handler.ts
pmelab: The message is not an issue, since its hardcoded within the application. But the destination comes from url or headers.
For login links i could encode destination into the JWT payload. And for logout, might just redirect to the login page instead.

Leksat: I think it's easier to escape them once rather than checking if the values are safe every time an eye catches an unescaped html template 🤷

pmelab:
@Leksat : it strips the tags from the redirect message now, and there are stricter checks on what can go into the destination url

.idea/prettier.xml
```diff
@@ -3,6 +3,6 @@
   <component name="PrettierConfiguration">
     <option name="myRunOnSave" value="true" />
     <option name="myRunOnReformat" value="true" />
-    <option name="myFilesPattern" value="{**/*,*}.{js,cjs,mjs,ts,jsx,tsx,gql,graphql,graphqls,md,mdx,json}" />
+    <option name="myFilesPattern" value="{**/*,*}.{js,cjs,mjs,ts,jsx,tsx,gql,graphql,graphqls,md,mdx,json,htm,html}" />
```
Leksat, line 6: Can be also added to https://github.com/AmazeeLabs/silverback-template/blob/6bfca4765c930a188d4b2a7e2edf0a627c052ba4/package.json#L15-L16

packages/token-auth-middleware/src/lib/email-backend.ts
```diff
@@ -0,0 +1,105 @@
...
+  ) {}
+
+  async getInfo(id: string): Promise<{ id: string; name: string } | undefined> {
+    if (!id) {
+      return;
+    }
+    if (id && this.users[id]) {
+      return { id, name: this.users[id] };
+    }
+    const [name, host] = id.split('@');
```
Leksat, line 28: Email addresses like `"@amazeelabs.com@"@evil.com` are actually valid.

I could not find email services that accept them. I also checked the PostMark service - it does not email to milti-@ addresses.

Yet, I think it's still better to split on the last `@` occurrence. Or first check if there is just a single `@` in the email.

packages/decap-cms-backend-token-auth/src/types.d.ts
```diff
@@ -0,0 +1,168 @@
...
+      auth: { status: boolean };
+      api: { status: boolean; statusPage: string };
+    }>;
+  }
+}
+
+declare module 'decap-cms-backend-github' {
+  import { Implementation } from 'decap-cms-lib-util';
+
+  export class GitHubBackend implements Implementation {
```
Leksat, line 93: So sad they don't exported types 😕

Leksat (changes requested):
Manual testing showed that the login link final destination is `https://slb-204.netlify.app/admin/_github` which shows some json. It worked well during the previous review.

packages/decap-cms-backend-token-auth/src/types.d.ts
pmelab: Its even written in typescript! All it needs is a `declaration: true` and a `types` in package.json. Maybe we can open a PR ...

Leksat: Would be nice. Else we might have a boom in production after a dependencies upgrade 😅

packages/token-auth-middleware/src/lib/email-backend.ts
pmelab: Done! Thanks for the catch!

Leksat (approved):
🎉 

Okay to merge once the tests are green 😛

PR #134, opened 2024-01-26
SLB-216: adopt flowbite components as base

luqmaanessop:
## Description of changes

Flowbitify starter components

PR #152, opened 2024-03-13
SLB-274 Shadcn UI - OLD

mattiasimonato:
## Description of changes
Integrate Shadcn into the system and establish an appropriate configuration and setup.

## Motivation and context
The final goal is to transition away from Headless UI in favor of adopting Shadcn within the project.

## How has this been tested?

- [x] Manually
- [ ] Unit tests
- [ ] Integration tests


pmelab (changes requested):
From a code and setup perspective this looks great!
But we have a lot of unrelated changes that probably result from conflicts with #134 . I think the best approach would be to reset the branch, rebase it and only commit what we actually need.

PR #174, opened 2024-04-10
SLB-218: use a simple inquiry webform for the mutation example

chindris:
## Description of changes
Instead of using the full fledged contact webform (which is pretty big now on the dev environment) for the mutation example, a new lighter webform is created, called inquiry. This webform is also exported to / imported from the config folder, so it is excluded from the config ignore settings.

## Related issue(s)
[Mutation example](https://amazeelabs.atlassian.net/browse/SLB-218)

## How has this been tested?
- [x] Manually
- [ ] Unit tests
- [x] Integration tests


pmelab (approved):
Perfect, thank you! 🥳

PR #175, opened 2024-04-10
SLB-146: silverback autosave and preview module

dspachos:
## Description of changes

SLB-146: silverback autosave and preview module

## How has this been tested?

- [ ] Manually


.lagoon/Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN npm install -g pnpm@8.6.12 && pnpm config set store-dir /tmp/cache/pnpm
 
 # Copy pnpm lockfile and install dependencies.
 COPY pnpm-lock.yaml .npmrc /app/
-#COPY patches /app/patches
+# COPY patches /app/patches
```
dan2k3k4, line 18: Why don't we commit a `.gitkeep` file in `/patches` directory and remove this comment so we never need to update this Dockerfile for projects which eventually get local (custom) patches?

apps/cms/composer.json
```diff
@@ -96,6 +96,9 @@
     "patches": {
       "drupal/userprotect": {
         "Fix site install": "https://www.drupal.org/files/issues/2023-07-28/3349663-8.patch"
+      },
+      "amazeelabs/silverback_gatsby": {
+        "Autosave preview": "./patches/fetch-entity-patch"
```
dan2k3k4: Should this file be a `.patch` file instead of dotless?

apps/preview/.eslintrc.json
```diff
@@ -0,0 +1,25 @@
+{
```
dan2k3k4, line 1: The eslintrc (`.eslintrc.json`) file format was deprecated in favour of flat config files, would require a migration to either one of these:
- eslint.config.js
- eslint.config.mjs
- eslint.config.cjs

https://eslint.org/docs/latest/use/configure/configuration-files

packages/drupal/silverback_autosave/README.md
```diff
@@ -0,0 +1,7 @@
+## Silverback autosave
+
+This module provides an autosave functionality on Guteneberg enabled nodes.
```
dan2k3k4:
```suggestion
This module provides an autosave functionality on Gutenberg enabled nodes.
```

packages/drupal/silverback_autosave/silverback_autosave.libraries.yml
```diff
@@ -0,0 +1,16 @@
+drupal.silverback_autosave:
+  version: VERSION
```
dan2k3k4: As @Leksat mentioned on Slack, we can drop this line

packages/drupal/silverback_autosave/silverback_autosave.module
```diff
@@ -0,0 +1,140 @@
...
+      ->get('silverback_autosave.settings')
+      ->get('allowed_content_entity_types');
+
+    // Autosave is enabled if either no restriction has been made or the current
+    // entity type is part of the restriction and as well either no restriction
+    // on bundle level has been made or the current bundle is part of that
+    // restriction.
+    if (!empty($allowed_content_entity_types)) {
+      if (!isset($allowed_content_entity_types[$entity_type_id]) ||
+        (!empty($allowed_content_entity_types[$entity_type_id]['bundles']) && !isset($allowed_content_entity_types[$entity_type_id]['bundles'][$entity->bundle()]))) {
```
dan2k3k4: This feels like it could be refactored to return early with single `if()` statements instead of multiple chained together

packages/drupal/silverback_autosave/silverback_autosave.module
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+}
+
+/**
+ * Implements hook_entity_update() for the user entity type.
+ *
+ * If the permissions of a user are changed, then we delete the autosave states
+ * belonging to that user. It is possible that new permissions the user is
+ * loosing the ability to access certain fields, but the autosave states contain
```
dan2k3k4:
```suggestion
 * losing the ability to access certain fields, but the autosave states contain
```

apps/preview/src/vite-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="vite/client" />
```
dan2k3k4, line 1: This file could be dropped in favour of adding `"vite/client"` to the `compilerOptions.types` of `tsconfig.json` - since we're currently not doing anything else special in this `vite-env.d.ts` file

https://vitejs.dev/guide/features#client-types

apps/cms/composer.json
dspachos: Correct, fixed.

packages/drupal/silverback_autosave/silverback_autosave.libraries.yml
dspachos: The point is correct, but instead of dropping I prefer to define the version (and this have full control over CDN in future updates)

PR #190, opened 2024-04-17
Gutenberg block: image with text

Leksat:
SLB-289

### DIFF
--- a/packages/ui/src/components/Organisms/PageDisplay.tsx
+++ b/packages/ui/src/components/Organisms/PageDisplay.tsx
@@ -43,6 +43,21 @@ export function PageDisplay(page: PageFragment) {
                     );
                   case 'BlockCta':
                     return <BlockCta key={index} {...block} />;
+                  case 'BlockImageWithText':
+                    return (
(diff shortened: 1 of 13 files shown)


packages/ui/src/components/Organisms/PageDisplay.tsx
```diff
@@ -43,6 +43,21 @@ export function PageDisplay(page: PageFragment) {
...
+                      <div
+                        style={{
+                          color: 'red',
+                          border: 'solid 3px red',
+                          padding: '3px',
+                          margin: '5px 0',
+                        }}
+                        // eslint-disable-next-line react/jsx-no-literals
+                      >
+                        BlockImageWithText goes here
```
dspachos, line 58: Surround with {'..'} perhaps? (and eliminate the `eslint` disable command)

Leksat: It errors even in this case 😅

Leksat: Still, it's just a stub.

dspachos (approved):

PR #192, opened 2024-04-17
chore: upgrade playwright and storybook

pmelab:
## Description of changes

* Upgrade Playwright and Storybook to the latest versions.
* Remove the dependency on `e2e` for installing playwright.

## Motivation and context

Resolve issues where `e2e` tests could not find playwright browser binaries.

## How has this been tested?

- [x] Manually
- [x] Unit tests
- [x] Integration tests


tests/e2e/package.json
```diff
@@ -7,6 +7,7 @@
   "license": "ISC",
   "type": "module",
   "scripts": {
+    "prep": "playwright install chromium",
```
Leksat, line 10: Hm. Previously there was a Playwright `postinstall` script which was downloading binaries during `pnpm i`. Looks like now it's gone.

Should we get rid of https://github.com/AmazeeLabs/silverback-template/blob/3414576a6e0ec0651fa1bb98c0cc1461699da12c/.lagoon/Dockerfile#L11 in this case?

Leksat (approved):
Looks great! Just one minor question.

tests/e2e/package.json
pmelab: Ah, is that why playwright is not downloading browsers automatically any more?

Leksat: Well, it's a long story. Previously, when it was called `playwright` (not `@playwright/test`) we had this issue with automatic browsers download in all environments. I went with `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1` and controlled downloads. Then when I faced the missing binaries error on this repo, I thought it's the same issue and added the env var. I did not know that the `postinstall` hook is already gone.

pmelab: Ok, thank you for the clarification. Will clean this up then.

Leksat: Aha! Now it's clear.

Playwright v1.38.0
> Breaking Changes: Playwright no longer downloads browsers automatically

Details: https://github.com/microsoft/playwright/releases/tag/v1.38.0

PR #224, opened 2024-05-08
fix: preview urls on dev and stage environments

colorfield:
## Description of changes

- add Lagoon `PROJECT_NAME` env var to preview
- adjust init
- refactor CMS env vars to use the same pattern and set it for dev environment

## Motivation and context

Noticed CORS issues on `dev` and `stage` when using the preview.
Domain prefix was `dev-.` 

See https://amazeelabs.slack.com/archives/C02ELKC86L9/p1715184263014659

## How has this been tested?

- [x] Manually
- [ ] Unit tests
- [ ] Integration tests


colorfield:
Tested and reviewed as part of #285, merging so we don't have a regression if we reset `stage`

PR #226, opened 2024-05-09
feat(SLB-306): Info Grid

elistone:
## Description of changes
Adds info grid block.

## How has this been tested?

- [x] Manually
- [x] Unit tests
- [ ] Integration tests


.idea/prettier.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="PrettierConfiguration">
+    <option name="myConfigurationMode" value="MANUAL" />
```
Leksat: What it does?

packages/drupal/gutenberg_blocks/src/blocks/info-grid-item.tsx
```diff
@@ -0,0 +1,63 @@
...
+registerBlockType('custom/info-grid-item', {
+  title: __('Info Grid Item'),
+  icon: 'align-wide',
+  category: 'layout',
+  attributes: {
+    icon: {
+      type: 'string',
+      default: '',
+    },
+  },
```
Leksat, line 24: 🐛 Shouldn't it be restricted to
```ts
  parent: ['custom/info-grid'],
```
?

Currently it can be added outside of `custom/info-grid`. Which does not fit the schema.

packages/drupal/gutenberg_blocks/src/utils/icon-list.ts
```diff
@@ -0,0 +1,130 @@
...
+      break;
+    case Icons.PHONE:
+      iconFileName = 'phone.svg';
+      break;
+    case Icons.LIFE_RING:
+      iconFileName = 'life-ring.svg';
+      break;
+  }
+
+  if (!iconFileName || iconFileName === '') {
```
Leksat: `!iconFileName` equals to `iconFileName === ''`. Something else was meant here?

packages/schema/src/schema.graphql
```diff
@@ -283,6 +284,16 @@ type BlockQuote @type(id: "custom/quote") {
   image: MediaImage @resolveEditorBlockMedia
 }
 
+type BlockInfoGrid @type(id: "custom/info-grid") {
+  items: [BlockInfoGridItem]! @resolveEditorBlockChildren
+}
+
+type BlockInfoGridItem @default @value {
```
Leksat: `BlockInfoGridItem` is never used as mandatory type, so `@default @value` can be removed.

packages/drupal/gutenberg_blocks/src/utils/icon-list.ts
```diff
@@ -0,0 +1,130 @@
...
+//
+// 1. Add a new icon to the Icons enum.
+// 2. Add a new icon to the allIconListOptions function.
+// 3. Add icon image to the ICON_IMAGE_PATH.
+// 4. Add a new icon to the iconImagePreview function.
+//
+// Your new icon is now available for selection, it will be displayed everywhere all icons are used.
+// Or you can use the limitedIconListOption function to limit the available icons in a specific context.
+//
+// NOTE: search this file for "*" to find the places where you need to add your new icon.
```
Leksat, line 14: I would simplify it. I think we can use TS types here in the way that no comments will be needed.

packages/drupal/gutenberg_blocks/src/utils/icon-list.ts
```diff
@@ -0,0 +1,130 @@
...
+    {
+      label: __('Life Ring'),
+      value: Icons.LIFE_RING,
+    },
+  ];
+
+  // If using default add to first place.
+  if (addDefault) {
+    allIcons.unshift({
+      label: __(defaultLabel),
```
Leksat: This does not work because it's `Drupal.t` under the hood. It can take only literal strings. Otherwise the string won't be extracted.

packages/drupal/gutenberg_blocks/src/utils/icon-list.ts
```diff
@@ -0,0 +1,130 @@
...
+ * @param defaultLabel
+ *  The default label.
+ * @param defaultValue
+ *  The default value.
+ */
+export const limitedIconListOption = (
+  icons: Icons[],
+  addDefault: boolean = true,
+  defaultLabel: string = 'Select an icon',
+  defaultValue: string = '',
```
Leksat, line 121: I would try to avoid adding a lot of parameters.
```ts
limitedIconListOption([...], true, 'Default', '_none')
```
is hard to understand.

If a function really needs a lot of params, I would make it a single object, so in the end you have
```ts
limitedIconListOption({
  icons: [...],
  addDefault: true,
  defaultLabel: 'Default',
  defaultValue: '_none',
})
```

More verbose, but easier to understand.

Leksat, line 121: At the moment the function is used with only the first arg passed. Everything else is always default.

It looks like the function is designed with future use-cases in mind. But that future might never come 😛

packages/drupal/gutenberg_blocks/src/blocks/info-grid-item.tsx
```diff
@@ -0,0 +1,63 @@
...
+              ])}
+              onChange={(icon: string) => {
+                setAttributes({ icon });
+              }}
+            />
+          </PanelBody>
+        </InspectorControls>
+
+        <div className={'container-wrapper'}>
+          <div className={'container-label'}>{__('Info Grid Item')}</div>
```
Leksat, line 48: Maybe just `Item`? Looks verbose to me at the moment
Screenshot 2024-05-16 at 15 21 21

But let Isa decided this 🙂

packages/drupal/gutenberg_blocks/src/blocks/info-grid-item.tsx
```diff
@@ -0,0 +1,63 @@
...
+  attributes: {
+    icon: {
+      type: 'string',
+      default: '',
+    },
+  },
+  edit: (props) => {
+    const { setAttributes } = props;
+    const iconPreview = iconImagePreview(props.attributes.icon as string);
+    const iconPreviewStyle = { maxWidth: '50px' };
```
Leksat: It is used just once. Maybe inline it?

packages/drupal/gutenberg_blocks/src/utils/icon-list.ts
Leksat: Actually I'm not sure if this file is needed at all. Because everything it contains is used only once, and maybe can be inlined (?)

packages/schema/src/schema.graphql
```diff
@@ -283,6 +284,16 @@ type BlockQuote @type(id: "custom/quote") {
...
+  icon: String! @resolveEditorBlockAttribute(key: "icon")
+  textContent: BlockMarkup @resolveEditorBlockChildren @seek(pos: 0)
+  cta: BlockCta @resolveEditorBlockChildren @seek(pos: 1)
```
Leksat: 🐛 From the grid item block source:
```tsx
          <InnerBlocks
            templateLock={false}
            allowedBlocks={['custom/heading', 'core/paragraph', 'custom/cta']}
```
This means that the block can have unlimited amount of children in any order. Tough, they are limited to 3 types.

Yet the schema here expects a fixed set of 3 blocks (2 first are grouped into 1).

I guess `templateLock` should be `true`?

packages/drupal/test_content/content/node/a397ca48-8fad-411e-8901-0eba2feb989c.yml
Leksat, line 1: Would it be possible to revert this file and create a new node with the new block? I think this one is overloaded 🙈

Leksat (changes requested):
@elistone There are just two comments which indicate probable bugs. Indicated with 🐛 

The rest are nit picks 🙂

packages/drupal/gutenberg_blocks/src/blocks/info-grid-item.tsx
isara89: No need to change - it's similar to `Accordion/Accordion item` :-)

elistone: Isa happy with how it is, will keep it the same.

.idea/prettier.xml
elistone: No idea, committed it by mistake I guess.

packages/drupal/gutenberg_blocks/src/utils/icon-list.ts
elistone: Well my idea here is that it should be used more that once, anywhere we give options to select icons in blocks.
I already can see in the CTA buttons we give the option for arrow icons before or after text but nothing gets displayed to the user while making these changes.

Also icon selection is quite a common request for PMs, so I was trying to make a reusable and dynamic libarary just for use on the Gutenberg side.

packages/drupal/test_content/content/node/a397ca48-8fad-411e-8901-0eba2feb989c.yml
elistone: So I can remove from here, but I noticed this file is what is used to create the tests for each block, so do we want to start adding more pages to the `blocks.sepc.ts` file?

Leksat (approved):

PR #238, opened 2024-05-15
Add "Conditional content" block

Leksat:
SLB-371

⚠️ If this is ever to be merged into the upstream, pick only relevant commits


Leksat:
Will be continued in https://github.com/AmazeeLabs/silverback-template/pull/310

PR #288, opened 2024-06-26
Add missing-translation redirect

Leksat:
SLB-423


Leksat:
Already merged as part of https://github.com/AmazeeLabs/silverback-template/pull/291

PR #289, opened 2024-06-27
Pre-Waku: RSC compatibility

pmelab:
## Description of changes

Switch to updated versions of various packages and adapt to incoming changes triggered by react server components.

## Motivation and context

First iteration of switching to an RSC based framework.

## How has this been tested?

- [x] Manually
- [x] Unit tests
- [x] Integration tests


apps/website/src/layouts/index.tsx
```diff
@@ -12,10 +12,14 @@ export default function Layout({
...
+      executors={[{ executor: drupalExecutor(`/graphql`) }]}
+    >
+      <OperationExecutorsProvider
+        executors={[{ executor: data, id: FrameQuery }]}
+      >
         <Frame>{children}</Frame>
-      </OperationExecutor>
-    </OperationExecutor>
+      </OperationExecutorsProvider>
+    </OperationExecutorsProvider>
```
Leksat, line 22: Can we use a single `OperationExecutorsProvider` passing an array of executors?

apps/website/src/utils/locale.ts
```diff
@@ -0,0 +1,33 @@
+// TODO: straight copy of the same file in the UI package
```
Leksat, line 1: Why the file from the UI package isn't used?

Leksat (approved):
Looks good!

Just few questions :)

apps/website/src/utils/locale.ts
pmelab: I did not want to create an outlier-dependency between ui and website that is only needed for Gatsby.

PR #296, opened 2024-07-04
SLB-445: Integrate Waku with Decap

pmelab:
## Description of changes

* Render Decap pages in Waku
* Use Decap translations in Waku

## Motivation and context

Use Decap contents in Waku. Also create an architecture that allows Waku to fetch data from multiple sources through the same GraphQL Schema.

## How has this been tested?

- [x] Manually
- [x] Unit tests
- [x] Integration tests


apps/website/src/entries.tsx
```diff
@@ -74,36 +88,34 @@ export default createPages(async ({ createPage, createLayout }) => {
   // Initialise a map for the homepages, since we want to exclude them from
   // creating a page for their internal path.
   const homePages = await query(HomePageQuery, {});
-  const homePageTranslations = new Map<Locale, Url>();
+  const homePageTranslations = [] as Array<Url>;
```
dan2k3k4: Why are we using `as` here?

```suggestion
  const homePageTranslations: Array<Url> = [];
```

apps/preview/server/index.ts
```diff
@@ -13,7 +13,7 @@ app.get('/endpoint.js', (_, res) => {
   res.send(
     `window.GRAPHQL_ENDPOINT = "${
       process.env.DRUPAL_URL || 'http://127.0.0.1:8888'
-    }/graphql";`,
+    }";`,
```
Leksat: I guess now `DRUPAL_URL` keeps the URL to the GraphQL endpoint? Should we rename the env var accordingly? Or is it `GRAPHQL_ENDPOINT` that should be renamed?

Leksat: Also, a suggestion:
```diff
- `window.GRAPHQL_ENDPOINT = "${var}";`
+ `window.GRAPHQL_ENDPOINT = ${JSON.stringify(var)};`
```

apps/cms/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    const headers = {
+      'SLB-Forwarded-Proto': publicUrl.protocol.slice(0, -1),
+      'SLB-Forwarded-Host': publicUrl.hostname,
+      'SLB-Forwarded-Port': publicUrl.port,
+      'X-Forwarded-Proto': publicUrl.protocol.slice(0, -1),
+      'X-Forwarded-Host': publicUrl.hostname,
+      'X-Forwarded-Port': publicUrl.port,
+    };
+    if (isMutation) {
+      const { data, errors } = await (
```
Leksat: I'd add try/catch.

apps/cms/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+            variables: variables || {},
+          }),
+          headers: {
+            ...headers,
+            'Content-Type': 'application/json',
+          },
+        })
+      ).json();
+      if (errors) {
+        throw errors;
```
Leksat: There are many reasons to throw `new Error(...)` (backtrace, compatibility with error handling of other libs, etc). Also, from personal experience, some libraries display only the first line of the error message if it's multi-line.

So maybe
```ts
const message = 'GraphQL returned errors';
console.error(message, errors);
throw new Error(message);
```
?

(Also https://typescript-eslint.io/rules/only-throw-error/)

Leksat: Also, according to the [spec](https://spec.graphql.org/October2021/#sec-Response-Format), the presence of `errors` does not mean that there is no `data`.

Maybe something like
```ts
if (data) {
  if (erros) {
    // log errors
  }
  // ...
} else {
  // log errors
  // throw error
}
```
?

apps/cms/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+      }
+      return data;
+    } else {
+      url.searchParams.set('queryId', id);
+      url.searchParams.set('variables', JSON.stringify(variables || {}));
+      const { data, errors } = await (
```
Leksat: Same things here.

apps/decap/src/collections/page.ts
```diff
@@ -166,64 +171,101 @@ const BlockMediaImageSchema = z
...
+                result = { ...translation };
+                result.translations = page.translations;
+              }
+            }
+          }
+        }
+        return result;
+      },
+    },
+  }) satisfies SourceResolvers;
```
Leksat, line 271: You could save some indent by moving `SourceResolvers` to the function return type 😛 

Or is this some secret technique? Let TS infer the function return type from the actual data making sure it satisfies the given type? Hmm 🤔

apps/decap/src/error.ts
Leksat, line 1: I can't find usages of this file. Some magic? Or forgotten?

apps/decap/src/graphql.ts
```diff
@@ -0,0 +1,81 @@
...
+  operation: TOperation,
+  variables: OperationVariables<TOperation>,
+) => Promise<OperationResult<TOperation>> {
+  const rawOperations = readFileSync(
+    `${basePath}/node_modules/@custom/schema/build/operations.json`,
+  ).toString();
+  const rawSchema = readFileSync(
+    `${basePath}/node_modules/@custom/schema/build/schema.graphql`,
+  ).toString();
+  const registry = merge.all(registries) as SourceResolvers;
```
Leksat:
```suggestion
  const registry = merge.all<SourceResolvers>(registries);
```

apps/decap/src/graphql.ts
```diff
@@ -0,0 +1,81 @@
...
+              src: source[info.fieldName].replace('/apps/decap', ''),
+            });
+          } else {
+            return source[info.fieldName];
+          }
+        }
+      },
+    });
+    if (!result.data) {
+      throw result.errors;
```
Leksat: Again, throwing a not-`Error` can cause problems with debugging.

apps/decap/src/image.ts
```diff
@@ -0,0 +1,19 @@
+import type { GraphQLFieldResolver } from 'graphql';
+import sizeOf from 'image-size';
+import { lookup } from 'mime-types';
+
+export const imageProps: GraphQLFieldResolver<string, any> = (source) => {
```
Leksat: Cannot find usages. Magic? Or forgotten?

apps/decap/src/image.ts
```diff
@@ -0,0 +1,19 @@
...
+  const relativeSource = source.substring(`/apps/decap`.length);
+  const dimensions = sizeOf(`node_modules/@custom/decap/${relativeSource}`);
+  const imageSrc = `${
+    process.env.GATSBY_PUBLIC_URL || 'node_modules/@custom/decap'
```
Leksat: Gatsby? 😁

apps/website/src/drupal-executor.tsx
```diff
@@ -1,85 +1,14 @@
...
-      ).json();
-      if (errors) {
-        throw errors;
-      }
-      return data;
-    }
-  };
-}
-
 export function DrupalExecutor({
```
Leksat: Can't find usage. Magic or forgotten?

apps/website/src/entries.tsx
```diff
@@ -74,36 +88,34 @@ export default createPages(async ({ createPage, createLayout }) => {
...
+        ) {
+          pagePaths.add(translation.path);
+        }
+      });
+    });
+  }
+
   createPage({
     render: 'static',
     path: '/[...path]',
```
Leksat, line 121: One line below:
```ts
staticPaths: [...pagePaths].map(...)
```
Why array destructing?

apps/website/src/executors-client.tsx
```diff
@@ -1,14 +1,20 @@
...
-
-export function ExecutorsClient({ children }: PropsWithChildren) {
+export function ClientExecutors({ children }: PropsWithChildren) {
   return (
-    <DrupalExecutor
-      url={import.meta.env.WAKU_PUBLIC_DRUPAL_URL || 'http://127.0.0.1:8888'}
+    <OperationExecutorsProvider
+      executors={[
+        {
+          executor: createDrupalExecutor(
```
Leksat, line 11: Just curious, with Decap backend only, can we have Content Hub at all?

packages/ui/src/components/Routes/Frame.tsx
```diff
@@ -14,17 +14,11 @@ function filterByLocale(locale: Locale) {
     str.language === locale;
 }
 
-function translationsMap(translatables: FrameQuery['stringTranslations']) {
+function translationsMap(
+  translatables: Exclude<FrameQuery['stringTranslations'], undefined>,
```
Leksat: A bit shorter way:
```suggestion
  translatables: Required<FrameQuery>['stringTranslations'],
```

packages/ui/src/components/Routes/Page.tsx
```diff
@@ -1,28 +1,38 @@
...
   const [loc] = useLocation();
-  return <PageWithData pathname={loc.pathname} />;
+  return (
+    <Operation
+      id={ViewPageQuery}
+      variables={{ pathname: loc.pathname }}
+      all={true}
+    >
+      {(result) => {
+        if (result.state === 'success') {
```
Leksat, line 19: Should we have `else { return <PageNotFound /> }` here?

packages/ui/src/components/Routes/Page.tsx
```diff
@@ -1,28 +1,38 @@
...
+          const page = result.data.filter((res) => !!res.page).pop()?.page;
+          if (page) {
+            const translations = Object.fromEntries(
+              page.translations
+                ?.filter(isTruthy)
```
Leksat, line 24: FYI: In TS >= 5.5 `.filter(item => !!item)` would be enough.
https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#inferred-type-predicates

Leksat (commented):
Some minor things only.

In general, looks awesome 💪  The data flow is much easier to track now 🎉

apps/decap/src/collections/page.ts
pmelab: Yes, satisfies makes sure the type matches `SourceResolvers` but if the actual type is more specific, it will not loose that information, while a return type definition does.

apps/website/src/executors-client.tsx
pmelab: Not right now, but maybe with https://amazeelabs.atlassian.net/browse/SLB-208.

apps/website/src/entries.tsx
pmelab: `pagePaths` is a set, and `[...pagePath]` turns it into an array. Is it that what you mean?

packages/ui/src/components/Routes/Page.tsx
pmelab: With static site generation this is more of an internal error case (page is in page listing, but path does not resolve), and should not make it into production. The 404 page is fully client rendered, and that would mean adding a lot of unnecessary assets to normal pages.

packages/ui/src/components/Routes/Page.tsx
pmelab: Nice! But consequently, we would remove `isTruthy` entirely then. And thats to far out of scope for this PR.

pmelab:
@Leksat The green test on the admin ui is actually a false negative. The path can be accessed now, but the UI itself is broken because its not compatible with React 19. Either we manage to have two React versions in the repository without breaking Waku, or Decap compatibility is fixed at some point. Either way, I would defer this to another issue and close this PR to not have it generate conflicts for too long.

apps/decap/src/error.ts
Leksat: @pmelab I don't know why this one is marked as outdated. My comment is still relevant.

apps/website/src/entries.tsx
Leksat: Aha!

Leksat (approved):

PR #297, opened 2024-07-04
SLB-435: redirects

chindris:


apps/website/package.json
```diff
@@ -27,14 +27,16 @@
...
     "vitest": "^1.1.1"
   },
   "scripts": {
     "test:static": "tsc --noEmit && eslint '**/*.{ts,tsx,js,jsx}' --ignore-path='./.gitignore'",
     "full-rebuild": "pnpm clean && pnpm build:waku",
     "start:drupal": "pnpm run --filter @custom/cms start",
     "build:drupal": "CLOUDINARY_CLOUDNAME=test pnpm start-test start:drupal 8888 build:waku",
-    "build:waku": "waku build",
+    "build:waku": "waku build && pnpm redirects",
+    "redirects": "tsx ./src/build/redirects.ts --dir=dist/public",
```
pmelab: ULTRA NITPICK! 🤓 
Maybe call this `build:redirects` to make clear this is part of the build process.

apps/cms/config/sync/language/de/views.view.redirect.yml
```diff
@@ -44,19 +44,19 @@ display:
             label: Statuscode
             group_items:
               1:
-                title: '300 Multiple Choices'
+                title: '300 Mehrere Auswahlmöglichkeiten'
```
pmelab: Drush export shenanigans? I would rather not include this in the diff. Lets keep it english by default.

pmelab (changes requested):

apps/cms/config/sync/language/de/views.view.redirect.yml
chindris: This should be removed now.

apps/website/package.json
chindris: This is now renamed, I also think it makes more sense to have it called "build:redirects"

pmelab (approved):

PR #307, opened 2024-07-10
feat(SLB-450): preview auth

colorfield:
## Description of changes

- Adds OAuth with Drupal as IDP
- Skip OAuth if there is a valid preview link token for a given entity
- Optional Basic auth middleware if we want more simplicity

## Motivation and context

Authentication for the Preview app.

Depends on https://github.com/AmazeeLabs/silverback-template/pull/301

## How has this been tested?

- [x] Manually
- [ ] Unit tests
- [ ] Integration tests


pmelab (approved):

pmelab (commented):
I suppose the preview integration tests have to be adjusted to actually log in?

pmelab (approved):

apps/preview/package.json
```diff
@@ -14,20 +14,29 @@
   "dependencies": {
     "@custom/schema": "workspace:*",
     "@custom/ui": "workspace:*",
+    "cookie-parser": "^1.4.6",
     "express": "^4.19.2",
+    "express-basic-auth": "^1.2.1",
```
dan2k3k4, line 19: flagging as this seems to no longer receive updates since 3 years, and there were mentions of a `v2` for breaking changes but no branch for that exists on the repo

colorfield: Thanks, `1.2.1` still seems to be the latest version. Shouldn't be too much of an issue though, especially because we are only using it as a fallback and favour OAuth / set it by default.

PR #314, opened 2024-07-19
feat(SLB:455): a11y enhancements

luqmaanessop:
## Description of changes

a11y enhancements to improve usability


packages/ui/src/components/Organisms/ContentHub.tsx
```diff
@@ -99,3 +69,42 @@ export function ContentHub({ pageSize = 10 }: { pageSize: number }) {
     </div>
   );
 }
+
+export const Card = (item: ContentHubResultItemFragment) => (
```
dan2k3k4: Nitpick: We don't need to `export` this, as it's only used in this file.

I would either remove `export` or move the `Card` function to its own component file (e.g. as a Molecule)

packages/ui/src/components/Organisms/ContentHub.tsx
```diff
@@ -70,41 +70,51 @@ export function ContentHub({ pageSize = 10 }: { pageSize: number }) {
...
-        <span className="invisible w-0 h-0 overflow-hidden">{item.title} </span>
-        Read more
-        <svg
-          className="rtl:rotate-180 w-3.5 h-3.5 ms-2"
-          aria-hidden="true"
-          xmlns="http://www.w3.org/2000/svg"
-          fill="none"
-          viewBox="0 0 14 10"
+const Card = (item: ContentHubResultItemFragment) => {
+  const id = item.title.toLowerCase().replace(/\s+/g, '-');
```
dan2k3k4: I would not use the title for the ID, but get it from the ID of the map function and inject into the component.

As the regex is only replacing spaces, so `A title with "quotes" or exclamation! marks` will end up like `a-title-with-"quotes"-or-exclamation!-marks`

packages/ui/src/components/Organisms/ContentHub.tsx
```diff
@@ -70,24 +70,28 @@ export function ContentHub({ pageSize = 10 }: { pageSize: number }) {
...
+  id,
+}: {
+  item: ContentHubResultItemFragment;
+  id: number;
+}) => {
   return (
     <article
-      aria-labelledby={id}
-      className="focus-within:outline focus-within:outline-2 focus-within:outline-offset-2 focus-within:outline-indigo-600 relative max-w-sm bg-white rounded-lg hover:shadow dark:bg-gray-800 dark:border-gray-700 overflow-hidden flex flex-col-reverse"
+      aria-labelledby={`${id}`}
```
dan2k3k4: mmm `id` is already a JavaScript string, so `aria-labelledby={id}` should work, no need to parse into a template literal

apps/cms/composer.json
```diff
@@ -105,6 +105,9 @@
       "drupal/gutenberg": {
         "Gutenberg enabled hook": "https://www.drupal.org/files/issues/2024-05-07/gutenberg_enabled_hook_3445677-2.patch",
         "Remove !important from sidebar": "./patches/gutenberg_remove-important-sidebar.patch"
+      },
+      "drupal/graphql": {
+        "Check if translation exists when loading an entity by its uuid": "./patches/graphql_load_by_uuid_translation_check.patch"
```
dan2k3k4, line 112: Is this a custom patch? Or does it exist on drupal.org ?

It would be good to follow the convention for Drupal patches:

`#<Drupal-Issue-Number> - <Drupal-Issue-Title>` such as few lines above would be changed from:

>`"Gutenberg enabled hook": "https://www.drupal.org/files/issues/2024-05-07/gutenberg_enabled_hook_3445677-2.patch",`

To:

>`"#3445677 - Gutenberg enabled hook": "https://www.drupal.org/files/issues/2024-05-07/gutenberg_enabled_hook_3445677-2.patch",`

It would also be nice if we can contribute back these patches to drupal.org in a new issue

packages/drupal/custom/custom.module
```diff
@@ -225,6 +225,12 @@ function custom_form_views_exposed_form_alter(&$form, FormStateInterface $form_s
     // Not applicable
     $form['langcode']['#options']['zxx'],
   );
+  // The excludeIds filter contains a list off UUIDs, and this might exceed the
```
dan2k3k4:
```suggestion
  // The `excludeIds` filter contains a list of UUIDs, and this might exceed the
```

packages/schema/src/fragments/Card.gql
```diff
@@ -1,6 +1,11 @@
-fragment ContentHubResultItem on ContentHubResultItem {
-  title
+fragment CardItem on CardItem {
```
dan2k3k4, line 1: Filename is `Card.gql` but first fragment is `CardItem` - is this for `Card` or `CardItem` ? Would be good to standardise on naming it one or the other ways. Unless this `Card.gql` will contain more fragments like `CardTitle` or something?

packages/ui/src/components/Organisms/Card.tsx
```diff
@@ -0,0 +1,65 @@
+import { useIntl } from '@amazeelabs/react-intl';
+import { CardItemFragment, Image, Link } from '@custom/schema';
+import React from 'react';
+
+export const CardItem = ({
```
dan2k3k4, line 5: File is named `Card.tsx` but we export a `CardItem`, would be good to name one or the other ways?

packages/ui/src/components/Organisms/PageContent/BlockTeaserList.tsx
```diff
@@ -0,0 +1,88 @@
...
+  props: BlockTeaserListFragment & { excludeIds: Array<string | undefined> },
+) {
+  const intl = useIntl();
+  const { data, isLoading } = useOperation(TeaserListQuery, {
+    locale: intl.locale as Locale,
+    args: queryString.stringify(
+      {
+        title: props.filters?.title,
+        pageSize: (props.filters?.limit || '0') as string,
+        // The excludeIds field should contain a regular expreession value, so
```
dan2k3k4:
```suggestion
        // The excludeIds field should contain a regular expression value, so
```

packages/drupal/gutenberg_blocks/src/blocks/teaser-list.tsx
```diff
@@ -0,0 +1,117 @@
...
+import {
+  PanelBody,
+  SelectControl,
+  TextControl,
+  ToggleControl,
+} from 'wordpress__components';
+
+// @ts-ignore
+const { t: __ } = Drupal;
+
```
dan2k3k4:
```suggestion
```

packages/drupal/gutenberg_blocks/src/blocks/teaser-list.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { registerBlockType } from 'wordpress__blocks';
+import {
+  PanelBody,
+  SelectControl,
+  TextControl,
+  ToggleControl,
+} from 'wordpress__components';
+
+// @ts-ignore
+const { setPlainTextAttribute } = silverbackGutenbergUtils;
```
dan2k3k4, line 11: We should be able to add typing for this in the `declare global` found in `packages/drupal/gutenberg_blocks/src/index.ts`

packages/ui/src/components/Organisms/PageContent/BlockTeaserList.tsx
```diff
@@ -0,0 +1,88 @@
...
+}
+
+export function BlockTeaserList(props: BlockTeaserListFragment) {
+  const staticIds: Array<string | undefined> = [];
+  return (
+    <div className="bg-white py-12 px-6 lg:px-8">
+      <div className="mx-auto max-w-6xl">
+        <ul className="my-8 grid md:grid-cols-2 lg:grid-cols-3 gap-4 lg:gap-8">
+          {props.staticContent?.map((teaserItem) => {
+            staticIds.push(getUUIDFromId(teaserItem?.content?.id || ''));
```
dan2k3k4, line 34: This feels a bit strange here, instead of before the `return` block, although it would mean calling `map` twice (once before `return` and once during the `return`), it would be computed before React tries to render and return any HTML so the `DynamicTeaserList` should only get rendered once and not updated on each `staticIds.push` event

packages/ui/src/components/Organisms/PageContent/BlockTeaserList.tsx
```diff
@@ -0,0 +1,88 @@
...
+          return teaserItem ? (
+            <li key={teaserItem.id}>
+              <CardItem readMoreText={props.buttonText} {...teaserItem} />
+            </li>
+          ) : null;
+        })}
+      </>
+    );
+  }
+  return null;
```
dan2k3k4, line 87: If `isLoading` is true, we don't display any loader, it could be something for improvement

packages/drupal/gutenberg_blocks/src/blocks/teaser-list.tsx
```diff
@@ -0,0 +1,114 @@
...
+      type: 'boolean',
+    },
+    limit: {
+      type: 'string',
+    },
+    titleFilter: {
+      typ: 'string',
+    },
+  },
+  edit: (props) => {
```
dan2k3k4: We could extract the attributes from the props

dan2k3k4 (approved):

PR #325, opened 2024-08-12
chore(SLB-449): serve on non Lagoon environments

colorfield:
## Motivation and context

See https://github.com/AmazeeLabs/silverback-mono/pull/1563

Netlify url is now used for the Build url instead of the Publisher url.
As not serving in the build container will make `build.json` (that provides the latest build id) not available.

See https://github.com/AmazeeLabs/silverback-mono/blob/4a2160d4c3e9298286af3eac50eabf06ff029abd/packages/composer/amazeelabs/silverback_gatsby/src/GatsbyBuildTrigger.php#L137

## How has this been tested?

- [x] Manually
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/apps/cms/scaffold/settings.php.append.txt
+++ b/apps/cms/scaffold/settings.php.append.txt
@@ -14,7 +14,11 @@ $config['silverback_external_preview.settings'] = [
 
 $config['graphql.graphql_servers.main']['schema_configuration']['directable']['build_webhook'] = $publisherUrl . '/___status/build';
 $config['graphql.graphql_servers.main']['schema_configuration']['directable']['update_webhook'] = $publisherUrl . '/___status/update';
-$config['graphql.graphql_servers.main']['schema_configuration']['directable']['build_url'] = $netlifyUrl;
+if (getenv('SB_ENVIRONMENT')) {
+  $config['graphql.graphql_servers.main']['schema_configuration']['directable']['build_url'] = $publisherUrl;
+} else {
+  $config['graphql.graphql_servers.main']['schema_configuration']['directable']['build_url'] = $netlifyUrl;
+}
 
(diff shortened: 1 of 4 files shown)


dan2k3k4 (approved):

apps/cms/scaffold/settings.php.append.txt
```diff
@@ -14,7 +14,7 @@ $config['silverback_external_preview.settings'] = [
 
 $config['graphql.graphql_servers.main']['schema_configuration']['directable']['build_webhook'] = $publisherUrl . '/___status/build';
 $config['graphql.graphql_servers.main']['schema_configuration']['directable']['update_webhook'] = $publisherUrl . '/___status/update';
-$config['graphql.graphql_servers.main']['schema_configuration']['directable']['build_url'] = $publisherUrl;
+$config['graphql.graphql_servers.main']['schema_configuration']['directable']['build_url'] = $netlifyUrl;
```
Leksat: @colorfield Is `build_url` this used on local? If yes, we might need a condition here.

colorfield: thanks for catching @Leksat it's the main settings and we don't have local override so adding the condition.

colorfield: As long as no env var override is set locally, then it wouldn't cause issue due to https://github.com/AmazeeLabs/silverback-template/blob/release/apps/cms/scaffold/settings.php.append.txt#L6-L7, that's probably why tests were green, but we shouldn't assume this so checking it here https://github.com/AmazeeLabs/silverback-template/pull/325/commits/1c98c0cbaba4631750b2f83f3bd22590a8ac2eb1

PR #330, opened 2024-08-18
Builds On Github

Leksat:
SLB-453

⚠️ Before merge:
- drop `temp` commits


.github/workflows/fe_build.yml
```diff
@@ -0,0 +1,81 @@
...
+
+      - name: Prepare
+        run: SKIP_DRUPAL_INSTALL=1 pnpm turbo:prep
+        env:
+          TURBO_API: 'http://127.0.0.1:9080'
+          TURBO_TOKEN: 'local'
+          TURBO_TEAM: 'local'
+
+      - name: Publisher
+        uses: AmazeeLabs/publisher-action@4371b4a90a9fe2c86248e76c0f428e39abf6b610
```
dan2k3k4, line 45: could be automated as tagged version

dan2k3k4 (approved):

apps/publisher/publisher.config.ts
```diff
@@ -0,0 +1,83 @@
...
+};
+
+export default defineConfig(
+  isLagoon
+    ? {
+        ...base,
+        mode: 'github-workflow',
+        publisherBaseUrl: `https://${process.env.SERVICE_NAME}.${process.env.LAGOON_ENVIRONMENT}.${process.env.LAGOON_PROJECT}.${process.env.LAGOON_KUBERNETES}`,
+        workflow: 'fe_build.yml',
+        repo: 'AmazeeLabs/silverback-template',
```
pmelab, line 32: This should be updated by `INIT.md`, or people will be very confused.

pmelab (changes requested):
Very nice! 🥳 
Just one small documentation remark.

apps/publisher/publisher.config.ts
Leksat: Good catch, thank you!

Although, `pnpx @amazeelabs/mzx run INIT.md` is broken. It does nothing at all. I will fix it later. (I would actually replace it with a TS file and run via `tsx`)

.github/workflows/fe_build.yml
Leksat: Yes, I even tried to use tags first. But it's so annoying to update them :)

Leksat:
Merged: https://github.com/AmazeeLabs/silverback-template/commit/bcb0d37e3d3b17b1d1a6e12e3ce5a2477b0d8653

PR #333, opened 2024-08-25
chore: remove compose on blocks not using it

dan2k3k4:


packages/drupal/gutenberg_blocks/js/blocks/accordion-item-text.tsx
```diff
@@ -7,11 +7,9 @@ import {
 } from 'wordpress__block-editor';
 import { registerBlockType } from 'wordpress__blocks';
 import { PanelBody, SelectControl } from 'wordpress__components';
-import { compose, withState } from 'wordpress__compose';
 
 const { t: __ } = Drupal;
 
-// @ts-ignore
```
Leksat, line 14: ❤️

packages/drupal/gutenberg_blocks/js/blocks/accordion-item-text.tsx
```diff
@@ -25,25 +23,35 @@ registerBlockType('custom/accordion-item-text', {
...
   },
-  // @ts-ignore
-  edit: compose(withState())((props) => {
-    const { attributes, setAttributes } = props;
+  edit: (props: {
+    attributes: {
+      icon: string;
+      title: string;
+    };
+    setAttributes: (attributes: { icon?: string; title?: string }) => void;
```
Leksat: As discussed in the call:
- New typing kills some of the original typing
- We can do instead:
```ts
registerBlockType<{ title?: string; icon?: string }>(...
```

packages/drupal/gutenberg_blocks/js/blocks/accordion-item-text.tsx
```diff
@@ -25,25 +23,35 @@ registerBlockType('custom/accordion-item-text', {
...
+      icon,
     });
 
     return (
       <Fragment>
         <InspectorControls>
           <PanelBody title={__('Block settings')}>
             <SelectControl
-              value={attributes.icon}
+              value={icon}
```
Leksat: The `attributes.` prefix can actually be helpful if the code gets bigger.

packages/drupal/gutenberg_blocks/js/blocks/accordion-item-text.tsx
```diff
@@ -25,25 +23,35 @@ registerBlockType('custom/accordion-item-text', {
...
     const icons = [
       { label: __('- Select an optional icon -'), value: '' },
       { label: __('Checkmark'), value: 'checkmark' },
       { label: __('Questionmark'), value: 'questionmark' },
       { label: __('Arrow'), value: 'arrow' },
     ];
+
     setAttributes({
-      icon: attributes.icon === undefined ? '' : attributes.icon,
+      icon,
```
Leksat: This might change the logic.

packages/drupal/gutenberg_blocks/js/index.ts
```diff
@@ -44,4 +45,12 @@ declare global {
       }>;
     };
   };
+
+  const wp: {
+    richText: any;
```
Leksat: If not too hard, could you add a bit more detailed typing?
Something like:
```ts
  const wp: {
    richText: { unregisterFormatType: (block: string) => void };
```

Leksat (changes requested):
Please separate commits for different things 😛

Leksat:
@dan2k3k4 And the codegen script please 🙏

packages/drupal/gutenberg_blocks/js/blocks/accordion-item-text.tsx
```diff
@@ -10,8 +10,10 @@ import { PanelBody, SelectControl } from 'wordpress__components';
 
 const { t: __ } = Drupal;
 
-// @ts-ignore
-registerBlockType('custom/accordion-item-text', {
+registerBlockType<{
+  title: string;
+  icon: string;
```
Leksat: Here and in other places where you added typings to `registerBlockType`:

Attributes that have no `default` value should be typed as optional.

For example
```ts
  title?: string;
  icon?: string;
```

Leksat (changes requested):

packages/drupal/gutenberg_blocks/js/blocks/teaser-list.tsx
```diff
@@ -27,15 +27,18 @@ registerBlockType<{
...
     buttonText: {
       type: 'string',
+      default: '',
     },
     contentHubEnabled: {
       type: 'boolean',
     },
     limit: {
-      type: 'string',
+      type: 'number',
```
Leksat, line 36: This is a breaking change I'm afraid. No idea how existing blocks will behave after the type change. Have you checked?

Leksat (changes requested):

packages/drupal/gutenberg_blocks/js/blocks/teaser-list.tsx
dan2k3k4: As tested locally:
- if the field for some reason had a text string like `test`, then the field gets set to the number `0`
- backend editor works as expected

I've put `test` in the Limit for one node on `prod` so we can see if it breaks that build when we merge this, but it should work as expected

Leksat (approved):

PR #335, opened 2024-09-04
SLB-354: RSC Images

pmelab:
## Description of changes

Introduce a new image server component that processes images automatically.
Uses [@amazeelabs/image][pkg] which should also be reviewed in the process.

## Motivation and context

Optimized images in Waku (or other RSC-Frameworks) and a simplified interface for adding responsive images to react components. More details can be found on the package [README][pkg], but in short, all that is necessary is to add the image with its maximum displayed width (and optional height) and an optional pair of focus coordinates. Fine-tuning `srcSet` and `sizes` is an exercise in futility, so the component will automatically create a sensible set of images for all devices.

## How has this been tested?

- [x] Manually
- [x] Unit tests
- [x] Integration tests

[pkg]: https://github.com/AmazeeLabs/silverback-mono/tree/development/packages/npm/%40amazeelabs/image


packages/ui/package.json
```diff
@@ -39,6 +39,7 @@
     "report": "mkdir -p coverage/storybook && nyc report --reporter=lcov -t coverage/storybook --report-dir coverage/storybook"
   },
   "dependencies": {
+    "@amazeelabs/image": "^1.3.0",
```
Leksat: As a user of the package, I have the following thoughts:
1. Maybe move `Image` property descriptions from readme to the TS type? So that they are visible in IDE on hover.
2. The `focus` prop of `Image` - maybe rename to something more specific? `focalPoint`, `cropFocus`? To me it currently looks like the HTML focus which is a totally different thing.

packages/schema/src/fragments/PageContent/BlockImageTeasers.gql
```diff
@@ -1,7 +1,7 @@
 fragment BlockImageTeasers on BlockImageTeasers {
   teasers {
     image {
-      source(width: 1536, sizes: [[768, 768], [1536, 1536]])
```
Leksat, line 4: Magic numbers no more 🎉

packages/ui/src/components/Routes/Frame.tsx
```diff
@@ -24,7 +25,12 @@ function translationsMap(
   );
 }
 
-export function Frame({ children }: PropsWithChildren) {
+export function Frame({
+  children,
+  alterSrc,
+}: PropsWithChildren<{
+  alterSrc?: (src: string) => string;
```
Leksat: Maybe
```ts
alterSrc?: ImageSettings['alterSrc'];
```
?

apps/website/src/utils.ts
```diff
@@ -1,2 +1,5 @@
 export const drupalUrl =
   process.env.PUBLIC_DRUPAL_URL || 'http://127.0.0.1:8888';
+
+export const frontendUrl =
+  process.env.WAKU_PUBLIC_URL || 'http://127.0.0.1:8000';
```
Leksat, line 5: I hope with devbox we will be able to get rid of `ENV_VAR || 'default'` in the code 🤞

Leksat (approved):
Looks amazing!

pmelab:
More feedback from a discussion with @jasymons and @HagerDakroury : Right now everything is turned into a JPG, but we need to retain .png's for transparency.

HagerDakroury (approved):
No further comments from my side. Looks great, really excited to use it!

PR #338, opened 2024-09-09
CI: try devbox

pmelab:
## Description of changes

Add a starting point for using [devbox](https://www.jetify.com/devbox/docs/) for local development and CI testing.

## Motivation and context

Consistent, reproducible versions of php, nodejs and pnpm per project.

## How has this been tested?

- [x] Manually
- [ ] Unit tests
- [x] Integration tests

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,7 @@
+# Automatically sets up your devbox environment whenever you cd into this
+# directory via our direnv integration:
+
+eval "$(devbox generate direnv --print-envrc)"
+
+# check out https://www.jetpack.io/devbox/docs/ide_configuration/direnv/
+# for more details
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -3,20 +3,20 @@ description: 'Common setup steps for GitHub Actions'
 runs:
   using: 'composite'
   steps:
-    - name: Setup Node.js
-      uses: actions/setup-node@v4
+    - name: Install devbox
+      uses: jetify-com/devbox-install-action@v0.9.0
       with:
-        node-version: 18
+        enable-cache: 'true'
+    - name: Load devbox via direnv
+      uses: HatsuneMiku3939/direnv-action@v1
 
-    - name: Setup PHP
-      uses: shivammathur/setup-php@v2
-      with:
-        php-version: '8.2'
-
-    - uses: pnpm/action-setup@v4
-      name: Install pnpm
-      with:
-        version: 8.6.12
+    - name: Check xdebug config
+      shell: bash
+      run: |
+        if ! php --info | grep 'xdebug.mode => develop => develop'; then
+          echo 'Error: xdebug.mode should not be set in CI. Have you committed xdebug config by accident? Check devbox.d/php/php.ini'
+          exit 1
+        fi
 
     - name: Get pnpm store directory
       id: pnpm-cache
--- a/devbox.d/php/php.ini
+++ b/devbox.d/php/php.ini
@@ -1,0 +1,10 @@
+# Lagoon has less memory, but we need more for development environments because
+# Drupal installation takes lots of RAM.
+memory_limit = 256M
+
+# Xdebug.
+# Uncomment the first line to enable xdebug. And do NOT commit the change.
+;xdebug.mode=debug
+xdebug.start_with_request=yes
+xdebug.output_dir=/tmp/xdebug
+xdebug.profiler_output_name = cachegrind.out.%u
--- a/devbox.json
+++ b/devbox.json
@@ -1,0 +1,24 @@
+{
+  "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.12.0/.schema/devbox.schema.json",
+  "env": {
+    "DEVBOX_COREPACK_ENABLED": "true",
+    "PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS": "true"
+  },
+  "packages": [
+    "nodejs@18.19.0",
+    "nodePackages.pnpm@8.6.12",
+    "php@8.2",
+    "php82Packages.composer@2",
+    "php82Extensions.xdebug@3"
+  ],
+  "shell": {
+    "scripts": {
+      "clean": [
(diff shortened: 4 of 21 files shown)


devbox.d/php/php-fpm.conf
Leksat, line 1: Is this file mandatory? Can we get rid of it? Or make it smaller?

devbox.d/php/php.ini
```diff
@@ -0,0 +1,6 @@
+[php]
+
+; Put your php.ini directives here. For the latest default php.ini file, see https://github.com/php/php-src/blob/master/php.ini-production
+
+; memory_limit = 128M
+; expose_php = Off
```
Leksat: I'd also love to have an example for xdebug (commented out as well)

devbox.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.12.0/.schema/devbox.schema.json",
+  "env": {
+    "DEVBOX_COREPACK_ENABLED": "true"
+  },
+  "packages": [
+    "nodejs@18.19.0",
+    "nodePackages.pnpm@8.6.12",
+    "php@8.2",
+    "php82Packages.composer@latest",
```
Leksat: I would not use `latest`.

UPD: Ah, there is a lock file. But still, I would prefer to explicitly set at least a major version.

process-compose.yml
Leksat, line 1: Django? 🤔

Leksat:
For me Playwright works locally via devbox. Yet `apps/cms/.envrc` kills the root `.envrc`. So we probably should get rid of https://github.com/AmazeeLabs/silverback-mono/blob/development/packages/composer/amazeelabs/silverback-cli/assets/.envrc

But it is so nice to have Composer executables on path 🤔 

**UPD:** ✅  Solved by https://github.com/AmazeeLabs/silverback-mono/commit/5915bf9eeee7b41a318099ce9715620c8e3bd9c6

Leksat:
We can also try this: https://github.com/jetify-com/devbox-install-action/pull/25/files#r1491637940

**UPD:** ✅ It did not work, but https://github.com/marketplace/actions/direnv-action did

Leksat:
PhpSotrm
- https://plugins.jetbrains.com/plugin/15285-direnv-integration seems to work, but it runs `.envrc` before every script run. This adds a little annoying delay.
- ...

.github/actions/setup/action.yml
```diff
@@ -3,25 +3,12 @@ description: 'Common setup steps for GitHub Actions'
...
-      with:
-        php-version: '8.2'
-
-    - uses: pnpm/action-setup@v4
-      name: Install pnpm
-      with:
-        version: 8.6.12
-
+    - name: Install devbox
+      uses: jetify-com/devbox-install-action@v0.9.0
```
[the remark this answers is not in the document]
Leksat: Actually, not a good idea. Developers may have different versions installed. And currently `devbox` does not have a built-in feature to enforce a specific version of itself directly within the project configuration.

devbox.json
```diff
@@ -0,0 +1,24 @@
+{
+  "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.12.0/.schema/devbox.schema.json",
+  "env": {
+    "DEVBOX_COREPACK_ENABLED": "true",
```
Leksat, line 4: Gonna remove corepack in a new PR. Sometimes it collides with the installed packages and then we get a wrong pnpm version.

Leksat: Hm. Removing it leads to another problem:
```
$ node -v
v18.19.0

$ pnpm -v
8.6.12

$ pnpm i
 ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)

Your Node version is incompatible with "/Users/alex/work/silverback-template".

Expected version: ^18.19.0
Got: v18.17.1
```

devbox.json
```diff
@@ -0,0 +1,24 @@
...
+    "PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS": "true"
+  },
+  "packages": [
+    "nodejs@18.19.0",
+    "nodePackages.pnpm@8.6.12",
```
Leksat, line 9: FYI: `nodePackages.pnpm` was removed in https://github.com/AmazeeLabs/silverback-template/pull/341/commits/4951fe41906b4a7819c3d772df1553a0aabe8004

With `DEVBOX_COREPACK_ENABLED`, the pnpm version should be defined in the root `package.json`:
```json
  "packageManager": "pnpm@8.15.8",
```

.github/actions/setup/action.yml
```diff
@@ -3,20 +3,20 @@ description: 'Common setup steps for GitHub Actions'
 runs:
   using: 'composite'
   steps:
-    - name: Setup Node.js
-      uses: actions/setup-node@v4
+    - name: Install devbox
+      uses: jetify-com/devbox-install-action@v0.9.0
       with:
-        node-version: 18
+        enable-cache: 'true'
```
Leksat, line 9: FYI: Full cache was enabled in https://github.com/AmazeeLabs/silverback-template/commit/dcafe47357f149b4ec75854d075c82d689f7fe88

devbox.d/php/php.ini
```diff
@@ -0,0 +1,10 @@
+# Lagoon has less memory, but we need more for development environments because
+# Drupal installation takes lots of RAM.
+memory_limit = 256M
+
+# Xdebug.
+# Uncomment the first line to enable xdebug. And do NOT commit the change.
+;xdebug.mode=debug
```
Leksat, line 7: FYI: The default xdebug mode has been changed in https://github.com/AmazeeLabs/silverback-template/commit/9093d75bcf808219e7d07e216642d2e631e3ae0e

.envrc
Leksat, line 1: ℹ️ If you use this PR as an example to add Devbox to other projects, it's better to use it as a list of files, but copy the actual file contents from `release` branch. Because we had lots of improvements after this PR.

PR #360, opened 2024-10-23
chore(SLB-459): share accordion heading level info with accordion items

dan2k3k4:


packages/drupal/gutenberg_blocks/js/blocks/accordion-item-text.tsx
```diff
@@ -26,6 +27,10 @@ registerBlockType<{
     icon: {
       type: 'string',
     },
+    headingLevel: {
+      type: 'string',
+      default: 'h3',
```
colorfield: We  shouldn't need to set a default value here, the parent is set to default `h2` https://github.com/AmazeeLabs/silverback-template/blob/fefc9e6ea57975940c17c54beb16aedbc61992d5/packages/drupal/gutenberg_blocks/js/blocks/accordion.tsx#L21

colorfield: As we are not exposing the child value in the schema, another way would be to consume parent attributes via `Context` https://developer.wordpress.org/block-editor/reference-guides/block-api/block-context/ - but haven't used it yet, just for info as current implementation works perfectly fine, and would probably be more relevant when we want to expose multiple attributes.

PR #376, opened 2024-11-26
SLB 495 Import content AI

dspachos:


apps/converter/htmlToMarkdown.js
```diff
@@ -0,0 +1,172 @@
...
+  const html = match ? match[1] : null;
+
+  if (html) {
+    // Create a new JSDOM instance and parse the HTML string
+    const dom = new JSDOM(html);
+    // Extract the <main> element content
+    let mainElement = dom.window.document.querySelector('main');
+    if (!mainElement) {
+      mainElement = dom.window.document.querySelector('article');
+    }
```
dan2k3k4:
```suggestion
    const mainElement = dom.window.document.querySelector('main, article');
```

If `<main>` element does not exist, then it would find `<article>`

But not sure what occurs if both exist or multiple `article` exist 🤔 

EDIT: it will only match on the first one it finds in the DOM tree...

```
<body id="body">
  <article>test</article>
</body>
```
Would return `<body>`

But
```
<article>
  <body id="body">test</body>
</article>
```

Would return `article`

apps/converter/htmlToMarkdown.js
```diff
@@ -0,0 +1,172 @@
...
+function generateFolderName(content) {
+  const hash = crypto.createHash('md5').update(content).digest('hex');
+  return hash.substring(0, 12);
+}
+
+async function downloadImage(url) {
+  try {
+    const response = await fetch(url);
+    if (!response.ok)
+      throw new Error(`Failed to fetch image: ${response.statusText}`);
```
dan2k3k4:
```suggestion
    if (!response.ok) {
      throw new Error(`Failed to fetch image: ${response.statusText}`);
    }
```

apps/converter/htmlToMarkdown.js
dspachos: @dan2k3k4 all these are correct, this is a WIP, just want to deliver something for the upcoming webinar (with controlled input).

PR #377, opened 2024-11-26
fix: gutenberg_form_node_form_details_after_build() for $form['field_metatags']

colorfield:
## Description of changes

Do not unset `$form['field_metatags']` as `gutenberg_form_node_form_details_after_build()` will end up with

```Warning: Undefined array key "field_metatags" in gutenberg_form_node_form_details_after_build()```
```Warning: Trying to access array offset on null in gutenberg_form_node_form_details_after_build()```

We cannot simply just set `$form['field_metatags']['#access']` to `FALSE` as it will prevent to save metatags.

Changes

- Remove the specific condition that is limiting to the page CT with `node_page_form` and `node_page_edit_form`, in most projects we have several content types using Gutenberg and it's hard to discover that we need to adjust a condition
- Decouple the `moderation_state` and the `metatags` condition, as one can exist without the other
- Do not unset `$form['field_metatags']` but `$form['#fields_with_details']` key

## Motivation and context

Do not flood the logs and prevent to display these warnings in the node edit form when `$config['system.logging']['error_level'] = 'verbose';` is set.

## How has this been tested?

- [x] Manually
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/packages/drupal/custom/custom.module
+++ b/packages/drupal/custom/custom.module
@@ -14,6 +14,8 @@ use Drupal\media\Entity\Media;
 use Drupal\silverback_gutenberg\LinkProcessor;
 use Drupal\user\Entity\Role;
 use Drupal\user\UserInterface;
+use Drupal\Core\Entity\ContentEntityFormInterface;
+use Drupal\node\NodeInterface;
 use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
 
 /**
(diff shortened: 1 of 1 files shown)


dan2k3k4 (approved):

colorfield:
Found out that it will prevent to actually save the metatags, so going to refactor it a bit.

dan2k3k4 (approved):

PR #391, opened 2025-01-02
feat(SLB-512) Taxonomy filters content hub

elistone:
## Description of changes
Adding and exposing taxonomy filters onto the content hub.

## Motivation and context
To create better filtering for content.

## How has this been tested?

- [ ] Manually
- [ ] Unit tests
- [ ] Integration tests

Should be resolving #54


dan2k3k4 (approved):

packages/schema/src/operations/ContentHubTerms.gql
```diff
@@ -0,0 +1,8 @@
+query ContentHubTerms {
```
pmelab, line 1: By adding this one as a static query and injecting it as an executor, the options could be fetched at build time and appear immediately when the page is opened. It should work the same way as the `FrameQuery` in `layout.tsx`: 

https://github.com/AmazeeLabs/silverback-template/blob/288a886da9f917c4c3398af38774e76c57879006/apps/website/src/layouts/index.tsx#L13-L18

packages/schema/src/schema.graphql
```diff
@@ -451,6 +501,9 @@ type Query {
 
   stringTranslations: [TranslatableString!]
     @gatsbyNodes(type: "TranslatableString")
+
+  contentHubTerms: TermContentHubResults!
+    @loadByVocabulary(bundle: "content_hub")
```
pmelab, line 503: To make the above work, this would have to use `@gatsbyNodes` as well, so Gatsby knows about content hub terms.

packages/schema/src/schema.graphql
```diff
@@ -108,6 +113,46 @@ type WebsiteSettings
...
+  @value(json: "null") {
+  termId: String! @resolveProperty(path: "tid.value")
+  label: String! @resolveProperty(path: "name.value")
+  depth: Int!
+}
+
+"""
+Content Hub Terms Results from query
+"""
+type TermContentHubResults
```
pmelab: Is this type required? Couldn't the `Query.contentHubTerms` field directly output `[TermContentHub]`.

packages/drupal/custom/src/Plugin/GraphQL/Directive/TermsLoadByVocabulary.php
```diff
@@ -0,0 +1,52 @@
...
+      $storage = \Drupal::entityTypeManager()->getStorage('taxonomy_term');
+      $terms = $storage->loadByProperties(['vid' => $bundle]);
+      if ($terms) {
+        foreach ($terms as $key => $term) {
+          $items[$key] = $term;
+        }
+      }
+
+      return [
+        'total' => count($items),
```
pmelab: If the other changes are applied, we would also emit just the list of items directly, and no wrapping result opject.

pmelab (changes requested):
That works pretty well already!
I added some suggestions on how to make the select box appear immediately.

packages/ui/src/components/Organisms/ContentHub.tsx
```diff
@@ -35,18 +41,16 @@ export function ContentHub({ pageSize = 10 }: { pageSize: number }) {
     ),
   });
 
-  const termsResult = useOperation(ContentHubTermsQuery, {});
```
pmelab: You should be able to keep that `useOperation` as it is, no need to prop-drill values in here. You can just inject the static query into https://github.com/AmazeeLabs/silverback-template/blob/release/apps/website/src/layouts/index.tsx like the `FrameQuery`.

pmelab (changes requested):

packages/drupal/custom/src/Plugin/GraphQL/Directive/TermsLoadByVocabulary.php
```diff
@@ -0,0 +1,49 @@
...
+
+      $storage = \Drupal::entityTypeManager()->getStorage('taxonomy_term');
+      $terms = $storage->loadByProperties(['vid' => $bundle]);
+      if ($terms) {
+        foreach ($terms as $key => $term) {
+          $items[$key] = $term;
+        }
+      }
+
+      return $items ?: NULL;
```
pmelab, line 56: I'm so sorry I did not see this before 🙈 
This is missing cache information. So the result won't update if the terms change.
Another approach would be to configure a view that lists these terms and use the `@drupalView` directive instead.

packages/ui/src/components/Organisms/ContentHub.tsx
```diff
@@ -1,17 +1,19 @@
...
 
 import { isTruthy } from '../../utils/isTruthy';
 import { useOperation } from '../../utils/operation';
+import { Alert } from '../Molecules/Alert';
 import { Pagination, useCurrentPage } from '../Molecules/Pagination';
 import { SearchForm, useSearchParameters } from '../Molecules/SearchForm';
 import { Loading } from '../Routes/Loading';
 import { CardItem } from './Card';
 
 export type ContentHubQueryArgs = {
```
pmelab, line 14: Not caused by this PR, but it would be better to use `OperationVariables<typeof ContentHubQuery>` here to infer arguments from the query definition directly.

packages/ui/src/components/Molecules/SearchForm.tsx
```diff
@@ -1,12 +1,13 @@
...
-import { useLocation } from '@custom/schema';
+import { TermContentHub, useLocation } from '@custom/schema';
 import { zodResolver } from '@hookform/resolvers/zod';
 import React from 'react';
 import { useForm } from 'react-hook-form';
 import { z } from 'zod';
 
 const formValueSchema = z.object({
   keyword: z.string().optional(),
+  terms: z.string().optional(),
```
pmelab, line 12: We could add a type constraint here to make sure the zod schema always matches `OperationVariables<typeof ContentHubQuery>`. I think it would look something like:
```typescript
const formValueSchema : z.ObjectType<OperationVariables<typeof ContentHubQuery>> = ...;
```

pmelab (changes requested):
Three more things that I missed before 😬

packages/ui/src/components/Organisms/ContentHub.tsx
elistone: I think this will be more complex and I'm not totally sure were to make the change.
But `ContentHubQuery` does not return the arguments.

```
export type ContentHubQueryVariables = Exact<{
  locale: Locale;
  args?: InputMaybe<Scalars['String']['input']>;
}>;
```

```
export type ContentHubQuery = { contentHub: { total: number, items: Array<{ id: string, path: Url, title: string, hero?: { lead?: string | undefined, headline: string } | undefined, teaserImage?: { alt: string, source: ImageSource } | undefined, terms: Array<{ label: string, termId: string, depth: number }> } | { id: string, path: Url, title: string, hero?: { lead?: string | undefined, headline: string } | undefined, teaserImage?: { alt: string, source: ImageSource } | undefined, terms: Array<{ label: string, termId: string, depth: number }> } | undefined> } };
```

I think that is because these are `args` that just get passed to the `view` but none of this is defined in the schema.

packages/ui/src/components/Molecules/SearchForm.tsx
elistone: Ignoring the fact that `ContentHubQuery` does not contain what we want, `z.ObjectType` also does not exist I tried asking AI about this but I could not find an answer that was very helpful.

packages/ui/src/components/Organisms/ContentHub.tsx
pmelab: You are absolutely right. In that case we need the type definition.

packages/ui/src/components/Molecules/SearchForm.tsx
pmelab: I had to try it myself to remember how it works :) See my next PR.
If you would review that one, I think we are good to merge!

PR #412, opened 2025-01-31
[Experiment] New block implemented by cursor

Leksat:
I tried the Composer feature of the Cursor editor.

Initial prompt:
> I have a task for you:
>> Create a new gutenberg block "PersonTeaser" that accepts an image, a name and a birthdate. It should also be added to the graphql schema and output in the frontend.
>
> Before starting the task, find the locations of all files which might be affected. Also, a list of files which can be used as examples.
>
> Ask me if you aren't sure.

It showed the relevant path, but thought that the block's `.ts` file is the frontend. So I added:
> By frontend I meant the UI package.

Then it found all the things and I asked:
> Okay, let's start the implementation

After that I simply clicked `Accept all`

---

Also interesting:


chindris:
This is pretty impressive! I'll give Cursor a try this week as well.

PR #424, opened 2025-03-18
Navigation in preview mode

chindris:
## Description of changes
Support for navigation in preview mode.

After the new development in Silverback mono we are now able to request the latest revision of a particular content, by using the fetchEntity directive. This allows us to have a preview instance of the website where we can actually navigate through different pages instead of having just a simple preview of a particular content (node).

One of the most important changes is that the preview routes are not the form of __preview/content_type?nid=123 , but they are just the real path of the page (for example : /en/imprint).

The current workflow for sharing and accessing a preview of a content is now:
- the editor creates a share preview link (as before)
- the share preview link has only the preview token as a query parameter
- when the preview route is accessed with a preview token, this gets stored in the cookie as well, so the preview app is able to get it and forward to the cms app with every request.
- the cms app has an authentication provider that checks for a preview token and authenticates the request if the preview token is valid.
- the preview user can just navigate on the preview website as they would do on the real one (all the content paths should work just as for the real website, the difference is that the content will be retrieved using the drupal executor which will get the most up to date revision of a particular content).

## Related issue(s)
[https://amazeelabs.atlassian.net/browse/SLB-552](https://amazeelabs.atlassian.net/browse/SLB-552)


README.md
```diff
@@ -362,3 +362,35 @@ Refer to the
...
+  - silverback_autosave: 1.3.4
+  - silverback_preview_link: 1.6.12
+  - silverback_gatsby: 3.7.11
+  - custom content_preview module copied from this repository.
+- Make sure you have defined in `settings.php` the
+  `silverback_external_preview.settings.preview_host` which shuld point to the
+  domain (and optionally port) of the preview app, for example:
+  `http://127.0.0.1:8001`
+- There should be a `preview` role which should have at least the
+  `Aaccess any content revision` and the `Fetch any autosaved entity`
```
colorfield: typo:  `Access`

README.md
```diff
@@ -362,3 +362,35 @@ Refer to the
...
+
+The entire preview system works like this:
+
+- Make sure you have at least the following versions for these modules:
+  - silverback_autosave: 1.3.4
+  - silverback_preview_link: 1.6.12
+  - silverback_gatsby: 3.7.11
+  - custom content_preview module copied from this repository.
+- Make sure you have defined in `settings.php` the
+  `silverback_external_preview.settings.preview_host` which shuld point to the
```
colorfield: typo: `should`

README.md
```diff
@@ -362,3 +362,35 @@ Refer to the
...
+  `http://127.0.0.1:8001`
+- There should be a `preview` role which should have at least the
+  `Aaccess any content revision` and the `Fetch any autosaved entity`
+  permissions. Make sure you also have the custom `content_preview` module
+  enabled.
+- There should be a user created, having the `preview` role, that will be set as
+  `Default preview user` at `/admin/config/content/silverback_preview_link`
+- To create a shareable preview link:
+  - Open a content edit form
+  - Click on the previw button which should open the preview iframe in the left
```
colorfield: typo: `preview`

colorfield (approved):

README.md
chindris: should be fixed now.

README.md
chindris: should be fixed now.

README.md
chindris: should be fixed now.

PR #450, opened 2025-05-14
pnpm 10

Leksat:
pnpm 10 has new deploy mode which works with `inject-workspace-packages=true`

Currently `pnpm deploy` ignores the existing lockfile. It does the full resolution step which
- Takes time
- Can install dependency versions other than listed in the lockfile

The new deploy solves both issues:
- There's no resolution step
- Dependency versions are taken from the lockfile

At the moment https://github.com/pnpm/pnpm/issues/9283 prevents us from switching to pnpm 10.
(There can be more blockers, but we don't know yet.)


Leksat:
Closing: pnpm 10 fast deploy blocked.

Critical blocker: https://github.com/pnpm/pnpm/issues/9837 - Turborepo bypasses pnpm's script hooks, so sync-injected-deps-after-scripts never fires. Injected deps stay stale after builds.

Additional issues:
  - No wildcard support for script patterns (must list all 9+ prep scripts manually)

There's no sense in updating to pnpm 10 now.

Decision: Stay on current pnpm.

PR #475, opened 2025-05-21
Per-Project lagoon vector database

Leksat:
## Motivation and context

ALGMTB-67

> The lagoon-chat environment of the silverback-template project right now uses an external vector database provided by amazee.ai. For management and scalability reasons, it should rather have a dedicated container.

## How has this been tested?

- [x] Manually (in Docker locally)
- [ ] Unit tests
- [ ] Integration tests

<details>
<summary>Locally it worked</summary>

Screenshot_2025-05-21_at_17_40_59

</details>

---

The general test failure is unrelated - coming from the base branch.

---

Deployment might need several Drupal config import attempts. There can be errors, but they are gone just by re-running `drush cim` few times.

### DIFF
--- a/.lagoon/pgvector/init.sql
+++ b/.lagoon/pgvector/init.sql
@@ -1,0 +1,13 @@
+-- Enable the pgvector extension so we can store vector embeddings
+CREATE EXTENSION IF NOT EXISTS vector;
+
+-- Create the table that Search API will use for storing embeddings
+CREATE TABLE IF NOT EXISTS search_api_vector (
+  id BIGSERIAL PRIMARY KEY,
+  content VARCHAR,
+  drupal_entity_id VARCHAR,
+  drupal_long_id VARCHAR,
+  server_id VARCHAR,
+  index_id VARCHAR,
+  embedding vector(1024)
+);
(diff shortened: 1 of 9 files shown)


.lagoon/pgvector/init.sql
```diff
@@ -0,0 +1,13 @@
...
+-- Create the table that Search API will use for storing embeddings
+CREATE TABLE IF NOT EXISTS search_api_vector (
+  id BIGSERIAL PRIMARY KEY,
+  content VARCHAR,
+  drupal_entity_id VARCHAR,
+  drupal_long_id VARCHAR,
+  server_id VARCHAR,
+  index_id VARCHAR,
+  embedding vector(1024)
+);
```
dspachos, line 13: @Leksat can you please remove this? I will handle this inside the module code, there is a service with a method to create collections (which also includes the proper table data). It's ok though to leave the init.sql file (even if empty).

Leksat: As discussed in the call (and later it was confirmed by Philipp), we will leave this table to be used with search_api. And for other services we can create new tables.

PR #551, opened 2025-07-04
chore: root level command cleanup

pmelab:
## Description of changes

- Make prep _not_ depend on test static, to omit static checks during deployments. They just cost time.
- Remove commands that do the same as "precommit" to make it more clear what should be used.

## Motivation and context

- Agents are confused by too many commands.
- Lagoon deployments fail because they try to run PHP static tests.


Leksat (approved):
Less commands to remember YAY 😁

PR #565, opened 2025-08-22
feat: add react router bridge

HagerDakroury:
## Description of changes
- Add a bridge for react router that exposes https://reactrouter.com/api/components/Link and https://reactrouter.com/api/hooks/useLocation 

## How has this been tested?

- [ ] Manually
- [ ] Unit tests
- [ ] Integration tests


packages/@amazeelabs/bridge-react-router/src/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const navigate = useNavigate();
+
+  return [
+    {
+      pathname: location.pathname,
+      search: location.search,
+      searchParams: new URLSearchParams(location.search),
+      hash: location.hash,
+    },
+    navigate,
```
HagerDakroury, line 24: I think some options like [`preventScrollReset`](https://reactrouter.com/api/hooks/useNavigate#summary) would be helpful to expose (for example updating search params but avoiding scrolling to the top each time). But from my understanding, the type is defined here https://github.com/AmazeeLabs/silverback-template/blob/6adbcd3a7396875d72e073c8b1201675b4e0674e/packages/%40amazeelabs/scalars/src/index.tsx#L140 so we'll need to make changes there which might not be the best idea?

What do you think? @Leksat

packages/@amazeelabs/bridge-react-router/src/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <ReactRouterLink {...props} to={href || '/'}>
+      {children}
+    </ReactRouterLink>
+  );
+};
+
+export const LocationProvider: LocationProviderType = ({ children }) => {
+  return <>{children}</>;
+};
```
HagerDakroury, line 38: @Leksat Some other APIs can be nice to expose. For example:
- https://reactrouter.com/api/hooks/useSearchParams (can replace some current use cases of `navigate`)
- https://reactrouter.com/api/components/PrefetchPageLinks
- https://reactrouter.com/api/components/NavLink

But that would require changes in `@amazeelabs/scalars` and `@amazeelabs/bridge-storybook`. Not sure if it's simpler than that? 🤔

packages/@amazeelabs/bridge-react-router/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "scripts": {
+    "prep": "rm -rf build && tsc",
+    "precommit": "pnpm precommit:fix && pnpm precommit:check && pnpm test:unit",
+    "precommit:fix": "eslint . --fix --quiet",
+    "precommit:check": "tsc --noEmit && eslint . --quiet",
+    "test:static": "tsc --noEmit && eslint . --quiet",
+    "test:unit": "vitest run --passWithNoTests"
+  },
+  "dependencies": {
+    "react-router": "^7.8.1"
```
Leksat: Let's change it to `^7`? Or we might get several versions of the same package installed. (In the other project we have it at `7.6.2`)

packages/@amazeelabs/bridge-react-router/package.json
```diff
@@ -0,0 +1,49 @@
...
+  },
+  "peerDependencies": {
+    "react": "^18.3.1"
+  },
+  "devDependencies": {
+    "@amazeelabs/bridge": "workspace:*",
```
Leksat, line 28: I was curious how does it work if the rest of the packages do live in the monorepo. Then I found they were moved to the template 😅 

Could you please remove migrated packages from the monorepo?

packages/@amazeelabs/bridge-react-router/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "precommit:fix": "eslint . --fix --quiet",
+    "precommit:check": "tsc --noEmit && eslint . --quiet",
+    "test:static": "tsc --noEmit && eslint . --quiet",
+    "test:unit": "vitest run --passWithNoTests"
+  },
+  "dependencies": {
+    "react-router": "^7.8.1"
+  },
+  "peerDependencies": {
+    "react": "^18.3.1"
```
Leksat: Let's make it `^18`?

packages/@amazeelabs/bridge-react-router/src/index.tsx
Leksat: If we are to add one, it should works for most of frameworks. And the current ones are Waku and RR7.

Waku has a similar thing: https://github.com/wakujs/waku/blob/dbe3723186827e4fdabe3d5b38db9be201d482f0/packages/waku/src/router/client.ts#L144-L150

By the way, the default Waku behavior is
> scroll on path change (not on searchParams change)

(Maybe RR7 works the same? Have you checked this?) 

So if we are to add an option, I imagine it to be something like this:
```ts
  /**
   * Scroll behavior on navigation
   * - `true`: always scroll
   * - `false`: never scroll
   * - `undefined`: framework's default behavior
   */
  shouldScroll?: boolean;
```

And we would need to add its support to Waku bridge.

packages/@amazeelabs/bridge-react-router/src/index.tsx
Leksat: I'd say let's go with minimum for now. And once Philipp is back, you could check with him.

packages/@amazeelabs/bridge-react-router/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "description": "Bridge implementation for react-router.",
+  "private": false,
+  "sideEffects": false,
+  "publishConfig": {
+    "access": "public"
+  },
+  "type": "module",
+  "main": "build/client.js",
+  "types": "build/dts/client.d.ts",
+  "exports": {
```
Leksat: `main`, `types` and `exports` currently lead to non-existing files.

I'm not sure if we need client/server exports for this package.

Leksat (commented):
Looks good in general! Few things to check and discuss.

packages/@amazeelabs/bridge-react-router/src/index.tsx
HagerDakroury: As discussed with @Leksat:
- Scroll logic in RR7 is mostly managed in https://reactrouter.com/api/components/ScrollRestoration#scrollrestoration
- For now: we won't add this option to link/navigate in the bridge and if a custom logic is needed in a project using RR7 then we can first explore options in `ScrollRestoration` like https://github.com/remix-run/react-router/blob/2df250e8100d308261e2c17a58840e0d31835eac/packages/react-router/lib/dom/lib.tsx#L1929

Leksat (approved):

